PR #2, opened 2016-02-03
[TOREE 249] Enable pip install

Lull3rSkat3r:
This commit enables the building, packaging and release of a pip package.
The package can be installed via `pip install toree`. This then enables the
kernel installation via `jupyter toree install`.


Makefile
```diff
@@ -70,16 +81,29 @@ test:
...
 dist: kernel/target/scala-2.10/$(ASSEMBLY_JAR) ${shell find ./etc/bin/*}
-	@mkdir -p dist/toree-kernel/bin dist/toree-kernel/lib
-	@cp -r etc/bin/* dist/toree-kernel/bin/.
-	@cp kernel/target/scala-2.10/$(ASSEMBLY_JAR) dist/toree-kernel/lib/.
-	@echo "VERSION: $(FULL_VERSION)" > $(VERSION_FILE)
+	@mkdir -p dist/toree/bin dist/toree/lib
+	@cp -r etc/bin/* dist/toree/bin/.
+	@cp kernel/target/scala-2.10/$(ASSEMBLY_JAR) dist/toree/lib/.
+	@echo "VERSION: $(VERSION)" > $(VERSION_FILE)
 	@echo "COMMIT: $(COMMIT)" >> $(VERSION_FILE)
```
chipsenkbeil: It would be a good idea to include the commit information into the build info of the kernel: https://github.com/apache/incubator-toree/blob/master/project/Build.scala#L181

etc/pip_install/setup.py
```diff
@@ -0,0 +1,63 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+import sys
+from setuptools import setup
+
+setup_args = dict(
+    name='toree',
```
chipsenkbeil: We discussed this already, but it seems odd to `pip install toree` instead of `pip install toree-kernel` since this is geared towards installing the Toree kernel binary.

lbustelo: I think we can stick with `toree` as the package for now.

Makefile
lbustelo: @chipsenkbeil Can you open this as a separate JIRA?

Vagrantfile
```diff
@@ -124,7 +124,7 @@ cat << EOF > /home/vagrant/.ipython/kernels/toree-kernel/kernel.json
     "display_name": "Spark 1.5.1 (Scala 2.10.4)",
```
lbustelo: We should change this `display_name` to `Toree`. Thoughts on whether we should add the version here?

chipsenkbeil: Maybe add 1.5.1+ instead of just 1.5.1?

Makefile
Lull3rSkat3r: I have included this in the pip package, but we should open a new jira for the scala build.

Vagrantfile
Lull3rSkat3r: I removed the version all together to not cause any confusion between the version of Toree and Spark. To know what version of Spark a Toree build is compatible with, users should reference our site.

Lull3rSkat3r:
@lbustelo, all good to go.

chipsenkbeil:
Just noticed that installing Toree this way, there are two assembly jars under `/usr/local/share/jupyter/kernels/`.

- toree-kernel-assembly-0.1.0.dev1-SNAPSHOT.jar

- toree-kernel-assembly-0.1.0.dev2-SNAPSHOT.jar

Are both of those supposed to appear? This was with a fresh install on a new machine. I ran `sudo pip install toree` followed by `sudo jupyter toree install`.

chipsenkbeil:
Also, if we are going to create the `kernel.json` for them, would it make sense to create one per language we support? You're installing everything after all.

Have labels like:
- Toree (Scala 2.10.4)
- Toree (PySpark)
- Toree (SparkR)
- Toree (SparkSQL)

I think it's a good idea since we've started to see a couple of folks use PySpark and it would give more presence to the other languages that the kernel offers, so people don't continue to think that the kernel only supports Scala.

lbustelo:
LGTM

gurvindersingh:
So currently running "jupyter toree install" only install one kernel (scala). @chipsenkbeil as you mentioned in comment, it would be nice to have it installed kernel for all the supported languages. 

In general, in your experience which is a better way to run pyspark or SparkR using Toree or directly using normal python and R interpreter. Also if you say Toree can you please explain reason.

PR #4, opened 2016-02-08
Update dependency downloader

chipsenkbeil:
Adds the following options to %AddDeps:

- `--repository <URL>` to include another repository when downloading
- `--trace` to print out download information
- `--verbose` to print out miscellaneous information

```scala
%AddDeps org.apache.spark spark-streaming-mqtt_2.10 1.5.2 --trace --transitive --repository https://extra-repo.com
```

These options will show up in the help menu. You can provide `--repository` more than once to include more than one additional repository.

Will ignore any malformed URL. Will try repositories provided by `--repository` first before default repositories.

### DIFF
--- a/kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
+++ b/kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
@@ -17,12 +17,13 @@
 
 package org.apache.toree.boot.layer
 
+import java.io.File
 import java.util
 import java.util.concurrent.ConcurrentHashMap
 
 import akka.actor.ActorRef
 import org.apache.toree.comm.{CommManager, KernelCommManager, CommRegistrar, CommStorage}
-import org.apache.toree.dependencies.{DependencyDownloader, IvyDependencyDownloader}
+import org.apache.toree.dependencies.{CoursierDependencyDownloader, DependencyDownloader, IvyDependencyDownloader}
 import org.apache.toree.global
 import org.apache.toree.interpreter._
 import org.apache.toree.kernel.api.{KernelLike, Kernel}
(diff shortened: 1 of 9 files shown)


lbustelo:
taking a look at this now

lbustelo:
Tried a couple of dependencies and different repositories and all worked fine. 

I saw that dependencies are getting installed under `/tmp/.ivy2/https/repo1.maven.org/maven2/io/netty/netty/3.6.3.Final/netty-3.6.3.Final.jar`. Is it fair to say that we still do not have dependency isolation with `AddDeps` between kernel instances?

lbustelo:
Potential JIRAs that can be resolved by this

- https://issues.apache.org/jira/browse/TOREE-184
- https://issues.apache.org/jira/browse/TOREE-169
- https://issues.apache.org/jira/browse/TOREE-246

kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
```diff
@@ -122,8 +123,12 @@ trait StandardComponentInitialization extends ComponentInitialization {
   }
 
   private def initializeDependencyDownloader(config: Config) = {
-    val dependencyDownloader = new IvyDependencyDownloader(
+    /*val dependencyDownloader = new IvyDependencyDownloader(
       "http://repo1.maven.org/maven2/", config.getString("ivy_local")
+    )*/
+    val dependencyDownloader = new CoursierDependencyDownloader
+    dependencyDownloader.setDownloadDirectory(
+      new File(config.getString("ivy_local"))
```
chipsenkbeil: If we change this to something else, we could avoid clashes between kernel instances. Or, if we set the default ivy_local to a temporary directory, that would work as well.

chipsenkbeil:
Do we want to look into this under the same PR? Or open another issue? https://issues.apache.org/jira/browse/TOREE-169?focusedCommentId=15104588&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15104588

chipsenkbeil:
FYI, I read that using Coursier for resolution/downloading over standard sbt has a large increase in performance. Might be worth adding the plugin for us to see if it helps with Toree dependency resolution/downloading, especially when running automated build/test.

lbustelo:
Lets handle the location of the deps in a separate PR. I opened https://issues.apache.org/jira/browse/TOREE-261. We should just go ahead and get that work done asap since this is fresh.

PR #7, opened 2016-02-23
Added functions to send rich display content to client

lbustelo:
This PR adds a couple of functions to the `kernel` object to enable sending rich content to the client. For now it is supporting HTML and Javascript, but also has a generic method to send content of any mime-type that the client can handle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ dist/
 .example-image
 .ensime
 .ensime_cache/
+
+Untitled*.ipynb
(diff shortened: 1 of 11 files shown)


lbustelo:
fixes https://issues.apache.org/jira/browse/TOREE-258

lbustelo:
Still need to add tests.

@chipsenkbeil Can you take a quick look? Most importantly, I had to add support to save off `_lastKernelMessage` in the Comm messages.

lbustelo:
Tests are in. removing the WIP status

chipsenkbeil:
@lbustelo looked it over and everything checks out to me.

lbustelo:
@chipsenkbeil so no issues with storing the comm request as parents?

chipsenkbeil:
@lbustelo, I don't think so. We just use that to allow you to keep track of a specific parent header when creating streaming messages. That's kept in a closure with the stream once started, so I don't think a comm message will hurt it. If you want to be safe, create another global that holds on to the last comm kernel message instead of execution request kernel message.

lbustelo:
Ok.. merged

PR #8, opened 2016-02-24
Added initial plugin implementation

chipsenkbeil:
Adds a plugin module (unit and integration tests added). @lbustelo @Lull3rSkat3r we should talk about what to do with this.


chipsenkbeil:
Adding one last feature I think is useful (specifying a "suggested" priority for plugins) and refactoring what is returned from firing events (a map of plugin/method to result instead of just a sequence of results).

chipsenkbeil:
@lbustelo I would say that this is the more important of the two PRs to merge first. It will be much faster to rebase the Coursier than this one.

lbustelo:
All looks good. merged

PR #9, opened 2016-02-24
Updated to use coursier for dependency resolution/downloading

chipsenkbeil:
Switches dependency resolution/downloading from sbt's default to coursier (what we use for %AddDeps). Much faster, can run in parallel, can work offline using cache (sbt fails often if not online), more easily sandbox dependencies between projects, no global locks (waiting on ~/.ivy2/.sbt.ivy.lock), blah blah blah.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: scala
 scala:
     - "2.10.4"
-
+env:
+    global:
+        - COURSIER_NO_TERM=1
+        - TOREE_RESOLUTION_VERBOSITY=0
 jdk:
     - oraclejdk7
     - openjdk7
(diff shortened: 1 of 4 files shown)


chipsenkbeil:
Will come back to this one later.

chipsenkbeil:
Tested via `make jupyter`, `make dist` and using Travis to run "bare bones."

lbustelo:
taking a look at this now

chipsenkbeil:
Added `TOREE_RESOLUTION_VERBOSITY` so we can reduce how much resolution/downloading stuff is logged. Should be much cleaner output for Travis (kept standard logging for normal builds, Travis uses 0, which is minimal logging).

Squashed and rebased.

lbustelo:
It does behave faster. It is great when using `%AddDeps` in the notebook. merged!

chipsenkbeil:
Closes https://issues.apache.org/jira/browse/TOREE-270

PR #10, opened 2016-02-24
[WIP] Add scala 2.11.7 support

chipsenkbeil:
Work in progress.

Got Toree client/kernel to compile under 2.11.7, not including tests.
Compiling via `sbt ++2.11.7 compile`.
Will update makefile once 2.11.7 fully works.

Some notes:

1. Changes in REPL class server (no longer used, now using RpcEnv, same HTTP server as SparkContext?)
    - https://github.com/apache/spark/commit/4a46b8859d3314b5b45a67cdc5c81fecb6e9e78c
    - This is for Spark 2.0, not 1.6+

2. Most of Spark-specific REPL classes are gone (now serializable friendly)
    - https://github.com/apache/spark/commit/11e22b74a080ea58fb9410b5cc6fa4c03f9198f2

3. Scala 2.11 fix to allow dynamic add jar
    - https://github.com/scala/scala/pull/4051

I've already applied the Scala 2.11 fix to let us keep addJar/addDeps, although I still think we need to add the lazy interpreter to Scala so we can add jars to the compiler classpath before it starts. That requires magics to not be evaluated inside the interpreter. Discussed this with Justin last year.

Have not yet tackled issues with REPL class server settings (no longer created inside interpreter). Not sure of the ramifications of Spark-specific REPL classes being gone. Assuming we can just use the stock IMain (which is what I did) since it is serializable-friendly.

Things to point out:

1. Have specific 2.10/2.11 implementation of magic class creation since it uses Scala reflection, which changed a little bit
    - This should go away once we switch to the plugin system since it uses Java reflection

2. Have specific 2.10/2.11 implementation of parts of Scala interpreter that use SparkIMain versus IMain
    - We need to clean up the interpreter interface as it has methods that do not make sense like getting the class server uri


larisasawyer:
+1 for scala 2.11

leifwalsh:
+1.  Any timeline for this, or anything we can help with to unblock it?

chipsenkbeil:
@leifwalsh, we're going to have to figure out how the REPL works for 2.11 in terms of how class files are shipped over to Spark clusters. Previously, this was done with an entirely separate HTTP server, but it looks like they might have merged that in to the SparkContext's class server for 2.11.

I'm sure that there are several tests that are broken by this and will need to have Scala 2.10 and Scala 2.11 versions in the same manner that this branch provides implementations of the Scala interpreter class for 2.10 and 2.11.

@lbustelo, did we want this done before the first binary release?

lbustelo:
I don't think this should impede our 0.1.0 release plan. 

Gino B.

leifwalsh:
@chipsenkbeil ok, sounds like more work than I thought.  Should we expect this will take more than a couple of weeks, and roll our internal stuff back to 2.10 in the meantime?

lbustelo:
Based on the comments from @chipsenkbeil, I would think that this would take several weeks and be a very delicate process. Lots of parts to test. I would roll back to 2.10.

PiyushMishra:
@chipsenkbeil  Should we wait for SparkIMain to be decoupled from class server . How can we get SparkContext and Spark Conf inside ScalaInterpreterSpecific 2.11 trait so that we can create Classserver and set its URI in spark configuration.

mariusvniekerk:
Seems that the default Scala version for spark 2.0 also moved to 2.11.

msemelman:
+1

asorianostratio:
Hi, I'm interested in using Apache Toree with a Spark 1.6.2 Scala 2.11 version. Could you catch me up on the state of this integration? Anything we can help with to achieve this?

lbustelo:
At this moment we are only planning in supporting 2.11 from spark 2 and up. 

Gino B.

mariusvniekerk:
We can probably close this PR

lbustelo:
#56 covered this.

PR #11, opened 2016-03-02
[TOREE 256]: Add installation of all interpreters via pip.

Lull3rSkat3r:
Enables all interpreters to be installed by default. Users can specify the interpreters they want by providing a comma seperated list:

```
toree install --interpreters=PySpark,SQL
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ clean: clean-dist
 	$(call RUN,$(ENV_OPTS) sbt clean)
 	rm -r `find . -name target -type d`
 
-.example-image: EXTRA_CMD?=pip install jupyter_declarativewidgets==0.4.0; jupyter declarativewidgets install --user; jupyter declarativewidgets activate; pip install jupyter_dashboards; jupyter dashboards install --user; jupyter dashboards activate; apt-get update; apt-get install --yes curl; curl --silent --location https://deb.nodesource.com/setup_0.12 | sudo bash -; apt-get install --yes nodejs; npm install -g bower;
+.example-image: EXTRA_CMD?=printf "deb http://cran.rstudio.com/bin/linux/debian jessie-cran3/" >> /etc/apt/sources.list; apt-key adv --keyserver keys.gnupg.net --recv-key 381BA480; apt-get update; pip install jupyter_declarativewidgets==0.4.0; jupyter declarativewidgets install --user; jupyter declarativewidgets activate; pip install jupyter_dashboards; jupyter dashboards install --user; jupyter dashboards activate; apt-get update; apt-get install --yes curl; curl --silent --location https://deb.nodesource.com/setup_0.12 | sudo bash -; apt-get install --yes nodejs r-base r-base-dev; npm install -g bower;
 .example-image:
 	@-docker rm -f examples_image
 	@docker run -it --user root --name examples_image \
@@ -145,6 +145,14 @@ release: pip-release bin-release
 		python setup.py register -r $(PYPI_REPO) && \
 		twine upload -r pypi toree-$(VERSION).tar.gz'
 
+define JUPYTER_COMMAND
+pip install toree-$(VERSION).tar.gz
+jupyter toree install --interpreters=PySpark,SQL,Scala,SparkR
+cd /srv/toree/etc/examples/notebooks
+jupyter notebook --ip=* --no-browser
+endef
+
+export JUPYTER_COMMAND
 jupyter: DOCKER_WORKDIR=/srv/toree/dist
 jupyter: .example-image pip-release
-	@$(DOCKER) -p 8888:8888 --user=root  $(EXAMPLE_IMAGE) bash -c	'pip install toree-$(VERSION).tar.gz && jupyter toree install && cd /srv/toree/etc/examples/notebooks && jupyter notebook --ip=* --no-browser'
+	@$(DOCKER) -p 8888:8888  --user=root  $(EXAMPLE_IMAGE) bash -c "$$JUPYTER_COMMAND"
--- a/etc/bin/run.sh
+++ b/etc/bin/run.sh
@@ -16,7 +16,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License
 #
-                                           ``
 PROG_HOME="$(cd "`dirname "$0"`"/..; pwd)"
 
 if [ -z "$SPARK_HOME" ]; then
@@ -31,7 +30,12 @@ KERNEL_ASSEMBLY=`(cd ${PROG_HOME}/lib; ls -1 toree-kernel-assembly-*.jar;)`
 # disable randomized hash for string in Python 3.3+
 export PYTHONHASHSEED=0
 TOREE_ASSEMBLY=${PROG_HOME}/lib/${KERNEL_ASSEMBLY}
-exec "$SPARK_HOME"/bin/spark-submit \
-  ${SPARK_OPTS} \
-  --driver-class-path ${TOREE_ASSEMBLY} \
-  --class org.apache.toree.Main ${TOREE_ASSEMBLY} "$@"
+# The SPARK_OPTS values during installation are stored in __TOREE_SPARK_OPTS__. This allows values to be specified during
+# install, but also during runtime. The runtime options take precedence over the install options.
+if [ "${SPARK_OPTS}" = "" ]
+then
+   SPARK_OPTS=${__TOREE_SPARK_OPTS__}
+fi
+
+SPARK_OPTS="--driver-class-path=\"${TOREE_ASSEMBLY}\" ${SPARK_OPTS}"
+eval exec "${SPARK_HOME}/bin/spark-submit" "${SPARK_OPTS}" --class org.apache.toree.Main "${TOREE_ASSEMBLY}" "$@"
--- a/etc/pip_install/toree/toreeapp.py
+++ b/etc/pip_install/toree/toreeapp.py
@@ -18,38 +18,25 @@
 import os.path
 import sys
 import json
-from traitlets import Unicode, Dict
+import shutil
+from traitlets import Unicode, Dict, Set
 from jupyter_client.kernelspecapp  import InstallKernelSpec
 from jupyter_core.application import base_aliases
 from traitlets.config.application import Application
 from toree._version import __version__, __commit__
+from jupyter_client.kernelspec import KernelSpec
 
-KERNEL_SPEC= {
-    'display_name': '',
-    'language_info': { 'name': 'scala' },
-    'argv': [
-        '',
-        '--profile',
-        '{connection_file}'
-    ],
-    'codemirror_mode': 'scala',
-    'env': {
-        'SPARK_OPTS': '',
-        'MAX_INTERPRETER_THREADS': '16',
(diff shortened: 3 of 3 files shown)


chipsenkbeil:
@Lull3rSkat3r, is this good to go? Do you want me to look over it and merge if ready? Or is there still something to be done?

Lull3rSkat3r:
I needed to rebase which I just did. Also, @lbustelo had some input, so wait for him.

Makefile
```diff
@@ -139,4 +139,4 @@ release: pip-release bin-release
 
 jupyter: DOCKER_WORKDIR=/srv/toree/dist
 jupyter: .example-image pip-release
-	@$(DOCKER) -p 8888:8888 --user=root  $(EXAMPLE_IMAGE) bash -c	'pip install toree-$(VERSION).tar.gz && jupyter toree install && cd /srv/toree/etc/examples/notebooks && jupyter notebook --ip=* --no-browser'
+	@$(DOCKER) -p 8888:8888 -p 5005:5005 --user=root  $(EXAMPLE_IMAGE) bash -c	'pip install toree-$(VERSION).tar.gz && jupyter toree install --interpreters=PySpark,SQL,Scala,SparkR && cd /srv/toree/etc/examples/notebooks && jupyter notebook --ip=* --no-browser'
```
lbustelo: Is port 5005 there to enable debug?

etc/bin/run.sh
```diff
@@ -30,7 +30,10 @@ KERNEL_ASSEMBLY=`(cd ${PROG_HOME}/lib; ls -1 toree-kernel-assembly-*.jar;)`
 
 # disable randomized hash for string in Python 3.3+
 export PYTHONHASHSEED=0
+# The SPARK_OPTS values during installation are stored in TOREE_SPARK_OPTS. This allows values to be specified during
+# install, but also during runtime.
+export SPARK_OPTS="${TOREE_SPARK_OPTS} ${SPARK_OPTS}"
```
lbustelo: I did not mean to do this pattern here, but rather in the kernel spec file.

etc/pip_install/toree/toreeapp.py
```diff
@@ -18,38 +18,25 @@
...
-    'display_name': '',
-    'language_info': { 'name': 'scala' },
-    'argv': [
-        '',
-        '--profile',
-        '{connection_file}'
-    ],
-    'codemirror_mode': 'scala',
-    'env': {
-        'SPARK_OPTS': '',
```
lbustelo: I was wondering if in this file we could do
```
'SPARK_OPTS': ''<what we get through the install script>:${SPARK_OPTS}"
```

That way users can add or replace settings through the env variable.

Makefile
Lull3rSkat3r: Yes.

etc/pip_install/toree/toreeapp.py
Lull3rSkat3r: See my comment above.

etc/bin/run.sh
Lull3rSkat3r: Putting things in the kernel spec, as you state below, does not work. When the env vars from the kernel spec are exported, the substitution does not happen and `SPARK_OPTS` would be literally `<command line args during install> ${SPARK_OPTS}`. I tried different approaches, in `run.sh`, with evaluating `SPARK_OPTS` to expand the variable, but at that point the variable is set and there would be no expansion to what the user exports. This ultimately coming down to how Jupyter is spawning the kernels. We would need to make a change there to evaluate and expand the variables.

lbustelo: Nah! Just wanted to make sure you gave that a try. So the only comment really for me is to somehow make the env var `TOREE_SPARK_OPTS` seem more like an internal thing. Not sure if adding `__` in front or something similar. I don't want to document this variable and it's only supported use is to exchange settings between `jupyter toree install` and `run.sh`. 

Thoughts?

Lull3rSkat3r: Sure, we can obfuscate it. Will update and push my changes.

Lull3rSkat3r:
@lbustelo, push is changed and all good to go.

etc/bin/run.sh
```diff
@@ -30,7 +30,10 @@ KERNEL_ASSEMBLY=`(cd ${PROG_HOME}/lib; ls -1 toree-kernel-assembly-*.jar;)`
 
 # disable randomized hash for string in Python 3.3+
 export PYTHONHASHSEED=0
+# The SPARK_OPTS values during installation are stored in TOREE_SPARK_OPTS. This allows values to be specified during
+# install, but also during runtime.
+export SPARK_OPTS="${__TOREE_SPARK_OPTS__} ${SPARK_OPTS}"
```
lbustelo: Don't you need to separate with':'?

etc/bin/run.sh
```diff
@@ -30,7 +30,10 @@ KERNEL_ASSEMBLY=`(cd ${PROG_HOME}/lib; ls -1 toree-kernel-assembly-*.jar;)`
...
 
 exec "$SPARK_HOME"/bin/spark-submit \
   ${SPARK_OPTS} \
```
lbustelo: Run some tests with setting --driver-java-options with multiple Java options. I was having problems setting that on SPARK_OPTS and the run.sh failing. I think we need to wrap it with quoted or some other bash-fu to get it to work. 

In my tests I was setting log4j levels and passing debug options.

etc/bin/run.sh
Lull3rSkat3r: No, that would be needed for setting PATH, but not command line args. I am assuming these will expand into actual command line args.

lbustelo: Doh! Sorry about that. Don't know what I was thinking about.

etc/bin/run.sh
Lull3rSkat3r: Yeah, so it looks like you are supposed to pass all the options in one flag, which we were not doing. However, the args are not in the right format and need some kind of coercing to pass them along. If I explicitly put the options in which are being passed, everything works as expected. 
Found [this](https://mail-archives.apache.org/mod_mbox/spark-dev/201404.mbox/%3CCABPQxsvpRzXyYbcjnPuvshO16V0pykdpjgs=NOdfT-FMd9Uwmw@mail.gmail.com%3E) mailing list discussion about spark submit and this problem. 
Not working yet, but it is  a step in the right direction.

Lull3rSkat3r:
I made some slight changes in how the `SPARK_OPTS` are merged. Instead of combining the values of `__TOREE_SPARK_OPTS__` and `SPARK_OPTS` they will be used in order of precedence. `SPARK_OPTS` will be used before `__TOREE_SPARK_OPTS__`. I did this to simplify how people will have to reason about ordering of their arguments, and to simplify `run.sh` :wink:.

I tested the following enumerations and I have recorded the behvior, which IMO is correct:

* This just starts the kernel with no extra arguments to spark
```bash
# Define no spark opts
jupyter toree install --interpreters=PySpark,SQL,Scala,SparkR
jupyter notebook --ip=* --no-browser'
```

* This starts the kernel and waits for a debugger to connect
```bash
# Define install spark_opts
jupyter toree install --spark_opts="--driver-java-options=-Dlog4j.logLevel=trace -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005" --interpreters=PySpark,SQL,Scala,SparkR
jupyter notebook --ip=* --no-browser'
```


* This starts the kernel and I only see warning logs
```bash
# Define install spark_opts and SPARK_OPTS
jupyter toree install --spark_opts="--driver-java-options=-Dlog4j.logLevel=trace -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005" --interpreters=PySpark,SQL,Scala,SparkR 
SPARK_OPTS="--driver-java-options=-Dlog4j.logLevel=warn" jupyter notebook --ip=* --no-browser'
```

* This starts the kernel and I only see warning logs
```bash
# Define only SPARK_OPTS
jupyter toree install --interpreters=PySpark,SQL,Scala,SparkR 
SPARK_OPTS="--driver-java-options=-Dlog4j.logLevel=warn" jupyter notebook --ip=* --no-browser'
```

There is a slight hack I had to do. For the case where there are no options passed I was receving an error. This was caused by an empty string being interpreted as the class file to load. I fixed this by adding in the `--driver-class-path` option @chipsenkbeil and I are adding for the magics work. I am open for a better solution, but am lacking the bash-fu to get there. 

Any feedback @lbustelo @chipsenkbeil ?

chipsenkbeil:
Spoke with you in person. Hack sounds fine since we'll always have that option of `--driver-class-path` in place.

lbustelo:
There may be a missed testcase which better resembles my docker run

```
# Define a multi JVM property --driver-java-options through the SPARK_OPTS env var
jupyter toree install --spark_opts="--driver-java-options=-Dlog4j.logLevel=warn" --interpreters=PySpark,SQL,Scala,SparkR 
SPARK_OPTS="--driver-java-options=-Dlog4j.logLevel=trace -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005" jupyter notebook --ip=* --no-browser'
```

Lull3rSkat3r:
Ok, last change should cover your use case, @lbustelo. You will need to specify the parameter like this for a command line arg:
```
--spark_opts='--driver-java-options="-Dlog4j.logLevel=trace -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005"'
```

Or like this for an environment variable:

```
SPARK_OPTS='--driver-java-options="-Dlog4j.logLevel=trace -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005"'
```

Lull3rSkat3r:
Branch is rebased and should is good to merge once someone reviews it.

lbustelo:
I'm reviewing

lbustelo:
I tested passing the multiple `--driver-java-options` as explained above. I used docker `-e` and it worked.

Now, on installing by default all the interpreters... I think that we should only install what we think is satisfactory working. Here are some of the issues that I've found when scala is not the default interpreter.

1. Magics are not handled at all. This means that from a SQL default interpreter I cannot add data to a context using any of the other languages supported. That makes the SQL interpreter very limited.

2. If we install PySpark and SparkR in a default run of `jupyter toree install` as stand-alone kernels, are we sure that they can satisfactory perform the majority of what a user could accomplish with the Python kernel in jupyter or IRkernel? If the answer is no, then I think that by default I would only install the Scala Kernel but keep the `--interpreter` flag to install the other kernels. If yes, then I guess I'm ok with. 

I will just offer that in the context of the `all-spark-notebook` docker image, I don't see us ever installing Toree as multiple kernels and only installing the Scala one. It will make selecting a notebook confusing when it comes to R and Python.

lbustelo:
Ran the streaming demo and all looked good.

Lull3rSkat3r:
@lbustelo, will rebase this and ping when it is ready.

lbustelo:
I tested the install of kernels and I think it looks good. It is the right choice to only install Scala by default. Let me know if you are done and I'll merge.

Lull3rSkat3r:
@lbustelo rebased and good to go!

PR #15, opened 2016-03-15
Adding support for sharing sqlContext across interpreters

lbustelo:
Fixes https://issues.apache.org/jira/browse/TOREE-166


lbustelo:
sharedsqlcontext

lbustelo:
@chipsenkbeil This is still WIP, but check out what I had to do to get this going. Considering these changes, I feel like we have many unnecessary methods in [Interpreter.scala](https://github.com/apache/incubator-toree/blob/master/kernel-api/src/main/scala/org/apache/toree/interpreter/Interpreter.scala). It seems that `bindSparkContext` and `bindSqlContext` will not be necessary, and rather something that should be API in `kernel` and Interpreters can get a hold during the `init` call.

chipsenkbeil:
@lbustelo, I agree completely with unnecessary methods in the interpreter. Initializing using the kernel object was a goal that Brian and I discussed and it looks like it makes more sense now, although you still have issues like lazily starting a SparkContext (and SQLContext) being bound to the interpreter later (more Scala than PySpark). Of course, if interpreters were made plugins, they could receive a SparkContext/SQLContext event when those were created and bind the variables at that point.

There are methods in the interpreter that are more interpreter-specific (class server uri stemming from the Scala 2.10 interpreter only), have a hideous API (interpret call's return value), or duplicate/confuse other methods (start vs init). Cleaning those up would be excellent.

lbustelo:
@chipsenkbeil Take a look at what I ended up doing in https://github.com/lbustelo/incubator-toree/blob/ShareSqlContext/kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala#L131. 

My goal was to surface the weird dependency between ScalaInterpreter and Kernel out to the ComponentInit level. That way, Kernel is pretty clean of the weirdness.

I also move `start` to be something that is call explicitly for ScalaInterpreter. Don't like how it ends up getting called, but I think that is now major scope creep for this PR. I have no issues with cleaning up the rest of the Interpreter interface and lifecycle as a separate PR.

lbustelo:
Thanks to @Lull3rSkat3r for spotting the missing line of code that got lost during my merge. All tests should pass now.

kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
```diff
@@ -72,25 +73,23 @@ trait StandardComponentInitialization extends ComponentInitialization {
...
-
     initializeSparkContext(config, kernel, appName)
 
+    interpreterManager.initializeInterpreters(kernel)
+
+    val responseMap = initializeResponseMap()
+
     (commStorage, commRegistrar, commManager,
-      manager.defaultInterpreter.orNull, kernel,
+      interpreterManager.defaultInterpreter.orNull, kernel,
```
chipsenkbeil: Is there a reason you did `orNull` here instead of just letting it fail if there is no default interpreter? We don't want the kernel to continue if there isn't a default interpreter, right?

kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
```diff
@@ -138,35 +137,47 @@ trait StandardComponentInitialization extends ComponentInitialization {
...
-    config: Config, interpreter: Interpreter,
+    config: Config, interpreterManager: InterpreterManager,
     dependencyDownloader: DependencyDownloader
   ) = {
     logger.debug("Constructing plugin manager")
     val pluginManager = new PluginManager()
 
     logger.debug("Building dependency map")
-    pluginManager.dependencyManager.add(interpreter)
+    pluginManager.dependencyManager.add(interpreterManager.interpreters.get("Scala").orNull)
```
chipsenkbeil: Similar to default interpreter question. Do we want to just pass null here? Or fail explicitly? Otherwise, having a missing Scala interpreter results in plugins down the line failing with a mysterious null pointer exception.

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
```diff
@@ -22,49 +22,58 @@ import java.net.{URL, URLClassLoader}
...
 import scala.tools.nsc.backend.JavaPlatform
-import scala.tools.nsc.interpreter.{OutputStream, IR, JPrintWriter, InputStream}
+import scala.tools.nsc.interpreter.{IR, InputStream, JPrintWriter, OutputStream}
 import scala.tools.nsc.io.AbstractFile
 import scala.tools.nsc.util.{ClassPath, MergedClassPath}
 import scala.tools.nsc.{Global, Settings, io}
 import scala.util.{Try => UtilTry}
 
-class ScalaInterpreter() extends Interpreter {
+class ScalaInterpreter(config:Config = ConfigFactory.empty) extends Interpreter {
```
chipsenkbeil: Probably want to make those `private val`. Otherwise, I think they are treated as `private var` by default. In case classes, this is different, of course.

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
```diff
@@ -228,16 +228,15 @@ class ScalaInterpreter() extends Interpreter {
...
-  protected def interpreterArgs(kernel: KernelLike): List[String] = {
+  protected def interpreterArgs(): List[String] = {
     import scala.collection.JavaConverters._
-    kernel.config.getStringList("interpreter_args").asScala.toList
-  }
-
-  protected def maxInterpreterThreads(kernel: KernelLike): Int = {
-    kernel.config.getInt("max_interpreter_threads")
+    if(config.hasPath("interpreter_args"))
+      config.getStringList("interpreter_args").asScala.toList
```
chipsenkbeil: Does `getStringList` return an empty list if the property doesn't exist? Or does it throw an error? If it returns an empty list, you can simplify this.

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
```diff
@@ -515,17 +514,12 @@ class ScalaInterpreter() extends Interpreter {
...
 
     doQuietly {
       logger.debug(s"Binding SparkContext into interpreter as $bindName")
-      bind(
-        bindName,
-        "org.apache.spark.SparkContext",
-        sparkContext,
-        List( """@transient""")
-      )
+      interpret(s"""def ${bindName}: org.apache.spark.SparkContext = kernel.sparkContext""")
```
chipsenkbeil: You used `classOf[SQLContext].getName` when binding the SQLContext this way. Do you want to change `org.apache.spark.SparkContext` to `classOf[SparkContext].getName` to be consistent?

kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
lbustelo: That was the code that was there already. I'm with you... maybe we should let it fail.

kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
lbustelo: Same as above... the code was already `val scalaInterpreter = manager.interpreters.get("Scala").orNull`

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
lbustelo: nope... it fails. And there is no API in `Config` to get Options or pass defaults. An approach I could take to clean this up is to do
```
config.withFallback(ConfigFactory.parse(Map("interpreter_args"->""))).getStringList("interpreter_args")
```

lbustelo: or maybe we should set `interpreter_args` to be empty default in reference.conf

lbustelo: it is already... https://github.com/apache/incubator-toree/blob/846292233c9f61bfc99579ec9a20a9d0ccfb3260/resources/compile/reference.conf#L43

I wonder what could be wrong here.

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
lbustelo: sure

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
chipsenkbeil: https://github.com/apache/incubator-toree/blob/master/kernel/src/main/scala/org/apache/toree/boot/CommandLineOptions.scala#L145

and

https://github.com/apache/incubator-toree/blob/master/kernel/src/main/scala/org/apache/toree/boot/CommandLineOptions.scala#L171

Maybe that is causing the array to be invalidated since not providing the commandline argument is setting the value to None?

kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
chipsenkbeil: Go for it.

kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
chipsenkbeil: I'd make the change like the comment above.

lbustelo:
@chipsenkbeil I think I addressed all your code comments. I'm going to finish up the example notebook and that would be the last part of this PR. Are there any other code comments?

chipsenkbeil:
LGTM

lbustelo:
Created https://issues.apache.org/jira/browse/TOREE-285 to contain the TODO for updating docs regarding the limitation due to use of defs.

lbustelo:
removing the WIP. this is ready to merge @chipsenkbeil or @Lull3rSkat3r

chipsenkbeil:
LGTM

PR #19, opened 2016-03-25
Converted magics to plugins

chipsenkbeil:
Fixed plugin test classpath to run script

Updated internal plugin loading to use only Toree-related classes

Updated compilation to use Java 1.6 to avoid issues on JDK 8

Fixed kernel hanging on shutdown


chipsenkbeil:
@lbustelo, we've gotten this to pass, but accidentally clobbered it. We really, really, REALLY need to raise the Akka timeout so we stop hitting these false failures.

Passing run of the tests: https://travis-ci.org/apache/incubator-toree/builds/118560171

Makefile
```diff
@@ -92,7 +93,7 @@ kernel/target/scala-2.10/$(ASSEMBLY_JAR): VM_WORKDIR=/src/toree-kernel
 kernel/target/scala-2.10/$(ASSEMBLY_JAR): ${shell find ./*/src/main/**/*}
 kernel/target/scala-2.10/$(ASSEMBLY_JAR): ${shell find ./*/build.sbt}
 kernel/target/scala-2.10/$(ASSEMBLY_JAR): project/build.properties project/Build.scala project/Common.scala project/plugins.sbt
-	$(call RUN,$(ENV_OPTS) sbt toree-kernel/assembly)
+	$(call RUN,$(ENV_OPTS) sbt -v toree-kernel/assembly)
```
lbustelo: Do you guys want to leave the `-v` here?

lbustelo:
Did you guys test the [brunel](https://github.com/Brunel-Visualization/Brunel) magic?

Lull3rSkat3r:
@lbustelo, we have not tried the brunel magic. Will try that and let you know.

Makefile
Lull3rSkat3r: We can remove this.

lbustelo:
@Lull3rSkat3r and I tested the magic and it did not work. This scenario is interesting because it is a plugin added using `%%AddJar`.

Lull3rSkat3r:
We are probably missing something in `AddJar` or our logic for backwards compatiability in `MagicManager` is flawed. Will look into it and let you know.

Lull3rSkat3r:
So the problem is the `AddJar` magic did not maintain the backwards compatibility of adding the jar to the classpath. The `findMagic` method in `MagicManager`  is also incorrect. This method is only called in the backwards compatible method but is searching through a list of plugins, which the old magics will not be. 

We will basically need to bring back the old `MagicLoader` class to maintain the compatibility. It might be easier to just mixin that behavior to `MagicManager` and remove it once we want to no longer maintain compatibility.

Another option is to just abandon the old compatibility and offer a migration guide from Spark Kernel to Toree. These changes would be minimal, similar to what was done in the magics for this PR ( e.g.[LsMagic](https://github.com/apache/incubator-toree/pull/19/files#diff-452e44f32062cc53a123077b54453ad7L23) ).

chipsenkbeil:
@Lull3rSkat3r we search against plugins because any magic compiled against Toree will now be a plugin. We should not be supporting magics compiled against a pre-release Toree. For Brunel, if the backwards compatibility works, it should be as simple as updating their dependencies to Toree 0.1.0.

The compatibility was meant to handle the case where they did not implement the plugin methods using annotations. All magics should now be loaded as plugins (when compiled against Toree 0.1.0), but old magics will still be executed directly instead of through the event messages.

Lull3rSkat3r:
Right, so I think we need to define more thoroughly what it means for a magic to be backwards compatible and what we are going to support. That is where I was getting confused. 

So here are the levels of backwards compatibility I see:
1. Magics being under the package 'com.ibm.spark.kernel'
2. Magics which are not plugins, but under 'org.apache.toree' (Brunel magic)

As @chipsenkbeil alluded above, the second item in the list is technically not a valid magic, since there has not yet been a release. Now the question remains about the first item.

IMO there is no backwards compatibility to support because all magics will and should need to be recompiled against the first release of toree because of the package changing. Since there is a binary incompatibility in that regard, I don't think there is much sense in supporting older styles. The only way this would make sense is if we were going to continue to look in the `com.ibm.spark.kernel` package for magics.

 I vote we put a PR into the Brunel magic repo and remove the backwards compatibility code in MagicManager. Thoughts @chipsenkbeil and @lbustelo?

lbustelo:
I'm fine with pushing a PR to Brunel.

When I think about backwards compat, I was not thinking about supporting the "com.ibm.spark" package. I was thinking that meant supporting the old way of authoring magic that required the class to be in a set package name, and that package name would be `org.apache.toree`. 

I'm now leaning to completely removing the back compat support. We have an opportunity here to simplify.

chipsenkbeil:
@lbustelo removed -v on sbt and got rid of backwards compatibility code.

lbustelo:
Ran streaming demo and all looked good. Will merge.

PR #21, opened 2016-03-30
Upgrading references to Spark to 1.6.1

lbustelo:
Addresses https://issues.apache.org/jira/browse/TOREE-267

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN curl -sL https://deb.nodesource.com/setup_0.12 | bash - && \
     npm install -g bower
 
 # for pyspark demos
-ENV APACHE_SPARK_VERSION 1.5.1
+ENV APACHE_SPARK_VERSION 1.6.1
 RUN apt-get -y update && \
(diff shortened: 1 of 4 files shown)


lbustelo:
This PR is now ready. 

To verify:
1. Run `make dev` and make sure it is on 1.6.0 (we are using all-spark-notebook and that is not yet at 1.6.1)
2. Run `make jupyter` same as 1
3. Run `make dev-binder` and see that the binder images is at 1.6.1 and examples work.

/cc @Lull3rSkat3r @chipsenkbeil

lbustelo:
Note... the binder image pulls in toree from pypi, so it is getting `0.1.0.dev4`. That package does not have the fix to know where py4j is.

To fully test binder, we will have to publish a `0.1.0.dev5`

chipsenkbeil:
@lbustelo, just ran `make jupyter` and printed out the version using `sc.version` and it still says `1.5.1` instead of `1.6.0` or `1.6.1`. Did I miss something? I saw Apache 1.6.1 was being used at the beginning when things were being built/run using sbt.

chipsenkbeil:
LGTM

PR #23, opened 2016-03-30
Fixed issue with rebinding defs into vals

lbustelo:
While working on #15, found out that `def` where getting redefined as `val` after doing an `%addJar`. This PR filters out Symbols that are `sourceMethods`.


lbustelo:
Worked with @chipsenkbeil.

@Lull3rSkat3r can you take a quick look and merge?

lbustelo:
This gist https://gist.github.com/lbustelo/16b58f21d2c6cd2b7ac352ea8a29fbda should show the issue without this pr.

chipsenkbeil:
LGTM (these tests really need to be fixed)

PR #24, opened 2016-03-31
TOREE-284 add support for custom SQL context class

kapil-malik:
**Today** Toree defines a SQL context object by first trying to load "org.apache.spark.sql.hive.HiveContext" using reflection. If it fails, it falls back to SQLContext.

**My usecase** I have a custom SQL context class, which I want toree to load by reading from a custom spark configuration property (spark.sql.context.class)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -130,9 +130,8 @@ cat << EOF > /home/vagrant/.ipython/kernels/toree-kernel/kernel.json
     ],
     "codemirror_mode": "scala",
     "env": {
-        "SPARK_OPTS": "--driver-java-options=-Xms1024M --driver-java-options=-Xmx4096M --driver-java-options=-Dlog4j.logLevel=trace",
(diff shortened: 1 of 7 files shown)


lbustelo:
@kapil-malik Thanks for the PR. My comments below mainly for discussion purposes.

I'm currently thinking that this type of option is more specific to Toree than to Spark. At this time, the codebase makes a distinction. Spark specific are in SparkConf, Toree specific are in a typesafe Config object. If we agree, then we need to modify the PR to read the value out of Config and then use `reference.conf` to hold the default.

Lull3rSkat3r:
@kapil-malik, any reason why you closed this?

kapil-malik:
@lbustelo thanks for the feedback. Yes, I realize it will be better off in Toree config instead of spark config. Will make the change.
@Lull3rSkat3r I reverted the changes to incorporate feedback by lbustelo above.

Will raise another PR for this issue soon. Thanks again.

PR #27, opened 2016-04-19
Updated kernel message relay to ignore unknown incoming and output messages

chipsenkbeil:
Resolves [TOREE-272](https://issues.apache.org/jira/browse/TOREE-272) in terms of the kernel freezing on unknown message types.

Tested using `jupyter notebook` and `jupyter console` with Jupyter 4.2.0. Before the fix, an exception would be thrown and the kernel would halt. Now, a warning is printed that the message of type "..." will be ignored.

/cc @lbustelo


chipsenkbeil:
Looks like the tests pass. The one error in one run was a timing issue (again).

lbustelo:
tested with Jupyter 4.2 and Ipywidgets 5.0 on master and replicated the error. Then tested with the changes here and the kernel started all the way.

PR #28, opened 2016-04-21
[TOREE-300] - Added comm info request handler

poplav:


poplav:
Resolves https://issues.apache.org/jira/browse/TOREE-300

poplav:
@lbustelo, rebased with master after your PR.  This should be ready for review.

kernel/src/main/scala/org/apache/toree/kernel/protocol/v5/handler/CommInfoRequestHandler.scala
```diff
@@ -0,0 +1,93 @@
...
+      }
+    }
+  }
+
+  override def process(kernelMessage: KernelMessage): Future[_] = future {
+    logKernelMessageAction("Initiating CommInfo request for", kernelMessage)
+
+    val commMap = (Json.parse(kernelMessage.contentString) \ "target_name").asOpt[String] match {
+      case Some(targetName) => {
+        if(targetName.equals("")) {
```
lbustelo: Does the targetName beed to be empty to mean ALL... or should we also return all the comm_is if the targetName is not in the message?

poplav: Yes, your right we should also return all comms if the target name is not in the message.  The [spec](http://jupyter-client.readthedocs.org/en/latest/messaging.html#comm-info) does not mention if the target name value is an empty string.  The fix is then to get rid of the if/else and just `buildCommMap(targetName)` and return all in the case of target_name missing where parsing the contentString for target_name will return none.

lbustelo:
looks good

PR #30, opened 2016-04-26
Modifying travis to run a make target that tests and audits the code

lbustelo:
The goal is to automate as much checking as possible on a PR. This will now run RAT as part of travis build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,10 @@ jdk:
     - openjdk7
 
 script:
-    - make test-travis
+    - make all-travis
+
+before_cache:
+    - make clean-travis
 
 sudo: false
 
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@
 # limitations under the License
 #
 
-.PHONY: help clean clean-dist build dev test test-travis release pip-release bin-release dev-binder .binder-image audit
+.PHONY: help clean clean-dist build dev test test-travis release pip-release bin-release dev-binder .binder-image audit audit-licenses
 
 VERSION?=0.1.0.dev6-incubating
 COMMIT=$(shell git rev-parse --short=12 --verify HEAD)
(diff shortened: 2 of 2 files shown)


lbustelo:
@Lull3rSkat3r Please check the results of Travis and if it looks like it did the test, RAT and clean, this should be good to merge.

lbustelo:
I'm going to close and reopen the PR to force another build. This will test if the after-script is doing its job.

lbustelo:
The rerun worked fine, I'll take it that it is cleaning up properly. There reason that I'm a bit sckeptical is that I see the caching happening before the `after_script`. So either it is caching the unlock, or this is just a problem that is needed under certain circumstances, such as a build failure.

@chipsenkbeil You remember the reason you had [this](https://github.com/apache/incubator-toree/blob/c3470855f8f7bf7a0cb03082d22fe1ef10f8da98/Makefile#L165)

chipsenkbeil:
@lbustelo that came from instructions in sbt regarding using Travis' caching: http://www.scala-sbt.org/0.13/docs/Travis-CI-with-sbt.html#%28Experimental%29+Reusing+Ivy+cache

lbustelo:
@chipsenkbeil Since I was able to run the PR twice, you think that the `after-script` is working?

lbustelo:
https://docs.travis-ci.com/user/caching/#Build-phases points to the cache taking place before `after-script`. What if I unlock in `before-script`?

lbustelo:
I think I'm going to switch the cleaning to happen on `before-cache` https://docs.travis-ci.com/user/caching/#before_cache-phase

lbustelo:
@chipsenkbeil I think the shell commands are basically the same. I moved the target to `before-cache` and it all seemed to work.

Either @chipsenkbeil  or @Lull3rSkat3r , merge this when you get a chance.

Lull3rSkat3r:
Merged in as commit https://github.com/apache/incubator-toree/commit/cea63655a47d55334492f532c8ce1a22b5a3a457

PR #32, opened 2016-04-28
Adding support for custom python exec (e.g. virtualenv)

ericchang:
Adds an optional kernel parameter PYTHON_EXEC

````
...
    "SPARK_HOME": "/usr/lib/spark",
    "PYTHON_EXEC" : "/usr/local/python/virtualenvs/myvenv/bin/python",
...
````

which, if unspecified, uses the previous hard-coded default of `python`.

https://issues.apache.org/jira/browse/TOREE-310

### DIFF
--- a/pyspark-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/pyspark/PySparkInterpreter.scala
+++ b/pyspark-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/pyspark/PySparkInterpreter.scala
@@ -32,11 +32,13 @@ import scala.tools.nsc.interpreter.{InputStream, OutputStream}
 /**
  * Represents an interpreter interface to PySpark. Requires a properly-set
  * SPARK_HOME, PYTHONPATH pointing to Spark's Python source, and py4j installed
- * where it is accessible to the Spark Kernel.
+ * where it is accessible to the Spark Kernel.  Optionally specify PYTHON_EXEC
+ * to override the default python executable "python'
  *
  */
 class PySparkInterpreter(
 ) extends Interpreter {
+  private val PythonExecEnv = "PYTHON_EXEC"
   private val logger = LoggerFactory.getLogger(this.getClass)
   private var _kernel:KernelLike = _
 
@@ -63,6 +65,7 @@ class PySparkInterpreter(
     )
 
   private lazy val pySparkService = new PySparkService(
+    Option(System.getenv(PythonExecEnv)).getOrElse("python"),
     gatewayServer,
     pySparkBridge,
     pySparkProcessHandler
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: pyspark-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/pyspark/PySparkInterpreter.scala [hunk 1]
```diff
@@ -63,6 +65,7 @@ class PySparkInterpreter(
     )
 
   private lazy val pySparkService = new PySparkService(
+    if (System.getenv(PythonExecEnv) == null) "python" else System.getenv(PythonExecEnv),
```
chipsenkbeil: FYI, you could do something like `Option(System.getenv(PythonExecEnv)).getOrElse("python")` instead.

pyspark-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/pyspark/PySparkInterpreter.scala
```diff
@@ -32,11 +32,13 @@ import scala.tools.nsc.interpreter.{InputStream, OutputStream}
...
  * Represents an interpreter interface to PySpark. Requires a properly-set
  * SPARK_HOME, PYTHONPATH pointing to Spark's Python source, and py4j installed
- * where it is accessible to the Spark Kernel.
+ * where it is accessible to the Spark Kernel.  Optionally specify PYTHON_EXEC
+ * to override the default python executable "python'
  *
  */
 class PySparkInterpreter(
 ) extends Interpreter {
+  val PythonExecEnv = "PYTHON_EXEC"
```
chipsenkbeil: Is there a reason that `PythonExecEnv` needs to be public instead of private?

pyspark-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/pyspark/PySparkInterpreter.scala
ericchang: Thanks for the suggestion, will do.

pyspark-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/pyspark/PySparkInterpreter.scala
ericchang: No reason, good catch.

ericchang:
@chipsenkbeil I've updated with your feedback.

chipsenkbeil:
@ericchang, sorry, but would you mind rebasing your PR? We're trying to keep our commit log clean(er). The only change is the merging of your other PR, so it should be pretty straight forward. Once you do that, I'll merge it in. :)

chipsenkbeil:
@Lull3rSkat3r, @lbustelo, if you look at the test results for this PR (https://travis-ci.org/apache/incubator-toree/builds/126418257), you'll notice that we still have two unstable tests in `TaskManager` and `BrokerTransformer`. Not sure what you want to do about them.

ericchang:
@chipsenkbeil no problem, I usually rebase just forgot that my other PR had been merged :)

chipsenkbeil:
Awesome, thanks!

PR #33, opened 2016-04-28
Remove the need to add assembly to driver-class-path

lbustelo:
This PR removes the need to have the assembly jar explicitly in the driver classpath. Now the `--driver-class-path` option is free for clients to use when installing Toree. It can be set through the `SPARK_OPTS`.


lbustelo:
Fixes https://issues.apache.org/jira/browse/TOREE-308

PR #35, opened 2016-05-03
Fix kernel json quote...

poplav:


lbustelo:
I'm merging this. Thanks @poplav

PR #36, opened 2016-05-03
Make add deps installation directory either a specified path or a unique temp path per toree process.

poplav:
Similarly to the logic of add jar this will:
* Allow the user to specify the dependency dir and use that if it exists, else will create a temp dir for that toree process.
* Update the magic-tutorial notebook to reflect the download dir location in the output.

### DIFF
--- a/kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
+++ b/kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
@@ -18,6 +18,7 @@
 package org.apache.toree.boot.layer
 
 import java.io.File
+import java.nio.file.{Paths, Files}
 import java.util.concurrent.ConcurrentHashMap
 
 import akka.actor.ActorRef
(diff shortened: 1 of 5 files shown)


poplav:
Resolves https://issues.apache.org/jira/browse/TOREE-261

kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
```diff
@@ -116,12 +117,17 @@ trait StandardComponentInitialization extends ComponentInitialization {
...
 
   private def initializeDependencyDownloader(config: Config) = {
-    /*val dependencyDownloader = new IvyDependencyDownloader(
-      "http://repo1.maven.org/maven2/", config.getString("ivy_local")
-    )*/
+    val depsDir = {
+      if(config.hasPath("deps_dir") && Files.exists(Paths.get(config.getString("deps_dir")))) {
+        config.getString("deps_dir")
+      } else {
+        Files.createTempDirectory("spark_kernel_add_deps").toFile.getAbsolutePath
```
lbustelo: maybe we should start using `toree_add_deps` instead of `spark_kernel`. Should change in addjar if necessary.

lbustelo:
other than the comments on name of the directory... the changes look good.

poplav:
@lbustelo updated the commits to reference toree and not spark_kernel (included the change to add_jar here)

lbustelo:
Ran meetup notebook and saw the folders get created for JARs and DEPS and saw content there after running cells. Also ran cells that used the JARs and the DEPS and they compile. That is good enough for me to merge this.

lbustelo:
The only thing that I'm going to fix on a separate commit straight to master is the name of the kernel in [here](https://github.com/apache/incubator-toree/pull/36/files#diff-5a58afa262c7d727edde469f8413626aR786).

@poplav We should make sure we save notebooks when running them using `make jupyter` not `make dev`. Right now the kernels are named different.

PR #38, opened 2016-05-05
Fix kernel info spec message structure due to spec changes.

poplav:
When working on [Migrate System Tests To Use Jupyter Kernel Test](https://issues.apache.org/jira/browse/TOREE-296?jql=project%20%3D%20TOREE) it was discovered that the [kernel info spec](http://jupyter-client.readthedocs.io/en/latest/messaging.html#kernel-info) changed.  
These changes are needed to make the Jupyter tests pass.

PR #39, opened 2016-05-09
Added test coverage using Jupyter Kernel Test

poplav:
Dependent on #38 for these tests to pass, will rebase after that is merged in.

This PR:
* Uses [jupyter_kernel_test](https://github.com/jupyter/jupyter_kernel_test), a generic testing client pluggable across all kernels.
* The tests are run by invoking the [jupyter kernel client](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/client.py) api from the [jupyter_client](https://github.com/jupyter/jupyter_client) project.
* Tests ensure kernel installed, kernel info, executing hello world, an sc parallelize operation, a magic operation via showtypes.
* The test framework includes other kernel client operations like inspect and complete, but toree does not support those at the moment so the output will show as skipped.
* The test framework limits its testing to the kernel client api, but we can build onto that if needed outside this pr.  This pr is limited to providing base coverage, similar to what the other kernel projects have running now.


poplav:
@lbustelo the travis issues/resolutions were (reflected in the latest commit):
* when pip packaging make sure we run the containers as root (pip permission issues on writing)
* I ran into the exact travis issue docuemented at [Buffer overflow in Java_java_net_Inet4AddressImpl_getLocalHostName of OpenJDK6 and OpenJDK 7](https://github.com/travis-ci/travis-ci/issues/5227) when introducing the docker service.  The possible resolutions are:
  * Run tests with `oraclejdk7` instead of `openjdk7`.  `openjdk7` is the one causing the issue.  I think this is best.
  * Implement the fix in the issue which modifies the /etc/hosts file in the travis `before_install`. /etc/hosts file changes on travis are:
screen shot 2016-05-10 at 10 28 22 am


* FYI, the build time is ~20 minutes in total now, where before it looked to be ~13 minutes.

.travis.yml
```diff
@@ -25,14 +25,24 @@ env:
         - TEST_DILATION=3
 jdk:
     - openjdk7
+    - oraclejdk7
```
lbustelo: Pick one or the other. We do not want to have a matrix build.

poplav:
This should be ready for review.

Lull3rSkat3r:
@poplav, it looks like this doesn't cover all the cases in [TOREE 296](https://issues.apache.org/jira/browse/TOREE-296). Can you comment on that JIRA what is not covered by this PR?

poplav:
@Lull3rSkat3r updated the JIRA.

chipsenkbeil:
@poplav, *sniff*, it's so beautiful to have real system tests. Brings a tear to the eye.

Also, you mentioned in the JIRA requesting that they open more APIs to be testable. Could you go ahead and do that for the Comm API at least since that's an important one?

poplav:
@chipsenkbeil , Yes that is an open item on [jupyter_kernel_test](https://github.com/jupyter/jupyter_kernel_test).  I haven't looked into detail yet how that would be implemented, but will..

Dockerfile.jupyter_kernel_tests
```diff
@@ -0,0 +1,32 @@
+#
```
lbustelo: I would move this file under `/etc/build/DockerFile.jupyter_kernel_tests` or rather than having to build the image... run the `jupyter/pyspark-notebook` directly and do all the steps in the CMD of the `docker run ... bash -c 'all steps here``.

Makefile
```diff
@@ -117,7 +117,7 @@ dev: .example-image dist
 			&& jupyter notebook --ip=* --no-browser"
 
 test: VM_WORKDIR=/src/toree-kernel
-test:
+test: jupyter-tests
```
lbustelo: don't combine test and system test... keep them separate. Leave `test` as is and create a `system-test` target.

Makefile
```diff
@@ -174,6 +174,17 @@ jupyter: .example-image pip-release
 	@$(DOCKER) -p 8888:8888  -e SPARK_OPTS="--master=local[4]" --user=root  $(EXAMPLE_IMAGE) bash -c "$$JUPYTER_COMMAND"
 
 ################################################################################
+# System Tests Using Jupyter Kernel Test (https://github.com/jupyter/jupyter_kernel_test)
+################################################################################
+jupyter-tests: pip-release
```
lbustelo: rename `jupyter-test` to `system-test`

lbustelo:
@poplav some more comments and then I think it should be ready to merge

lbustelo:
Add `system-test` to https://github.com/poplav/incubator-toree/blob/jupyter_kernel_tests/Makefile#L278

poplav:
@lbustelo comments should be resolved by the latest commit

Makefile
```diff
@@ -262,9 +275,9 @@ audit: sign audit-licenses
 
 publish: audit publish-bin publish-pip publish-src publish-jars
 
-all: clean test audit
+all: clean system-test test audit
```
lbustelo: Not sure if it matters, but if order of the dependencies is the same as order of run in make, then we should do `test` before `system-test`.

lbustelo: Did not quite understood the docs [here](http://www.gnu.org/software/make/manual/make.html#Prerequisite-Types)... but see what you figure out.

poplav: It looks like `system-test` which depends on `pip-release` compiles locally as does `test`.  From looking at the logs at test it looks like when it gets to `test` it eagerly compiles less (the test classes only).  If we move `test` before I think it will compile everything and when it gets to `pip-release` that part will be automatically skipped rather than the compilation process running partially in both targets.

See [here for test compilation build logs](https://travis-ci.org/apache/incubator-toree/builds/129268926#L1133)
See [here for pip release which does assembly which skips compiling test classes build logs](https://travis-ci.org/apache/incubator-toree/builds/129268926#L384)

I will reverse the order because that may make the compilation process better and intuitively it kind of makes more sense to detect unit test failures before getting to system tests.

lbustelo:
I tested running system test locally and works good. Will merge.

PR #40, opened 2016-05-10
Skip comment only lines in the scala interpreter.

poplav:
@chipsenkbeil does this fix https://issues.apache.org/jira/browse/TOREE-311?jql=project%20%3D%20TOREE?

### DIFF
--- a/scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
+++ b/scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
@@ -263,7 +263,7 @@ class ScalaInterpreter(private val config:Config = ConfigFactory.load) extends I
   override def interpret(code: String, silent: Boolean = false):
   (Results.Result, Either[ExecuteOutput, ExecuteFailure]) = {
     val starting = (Results.Success, Left(""))
-    interpretRec(code.trim.split("\n").toList, false, starting)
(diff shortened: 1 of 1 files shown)


chipsenkbeil:
We also need to consider multi-line comments (`/*`) if we're going to filter out that way. Might be a better idea to use regex to remove comments from the full code blob before splitting. E.g.

```scala
val x = 3 /*
some comment
*/ val y = 4
// another comment
z = 5 // final comment?
// no, this is the final comment
```

becomes

```scala
val x = 3
val y 
```

Would also be a good idea to have some sort of test (maybe system? @poplav?) that can verify that providing comments does not provide an empty syntax error like you've seen.

@Lull3rSkat3r, @lbustelo, feel free to chime in on this one with your own thoughts.

Lull3rSkat3r:
agreed 🎐

lbustelo:
I wonder is there is code in some Apache package to do this rather than rely on regex or string lookups directly.

lbustelo:
@poplav Can you rename this into a `[WIP]`? I think there is more work to do here.

gotoariel:
I'm not sure if this is the right place to comment, but having just switched to Toree it seems like this is just a symptom of the lack of :paste mode, along with having to use braces to give companion objects the same scope as their class, and not being able to do method chaining with dots at the beginnings of lines.

I know there are line magics, but I wonder if there's a way (maybe through a notebook magic) to make each cell be sent to the interpreter in :paste mode and disable the line-by-line functionality. This would bring things much more in line experientially with other REPLs.

PR #41, opened 2016-05-16
Updated loading of external plugins to check internal class information when recursing interfaces and super classes

chipsenkbeil:
Resolves [TOREE-314](https://issues.apache.org/jira/browse/TOREE-314).

@lbustelo, @poplav, we need to add a system test with a dummy external plugin that we can add to verify this works. For now, @tigerquoll and @danrope can verify by building and running against this version of Toree for Brunel. I've been getting an AbstractMethodError with Brunel since adding this fix (indicating that it is detected) instead of no error (not loading the plugin). I posted on Gitter the comparison between a Toree magic and the Brunel magic, which showed that the Brunel magic was not compiled against a new version of Toree (even when I built it myself using `gradle assemble`).

### DIFF
--- a/plugins/src/main/scala/org/apache/toree/plugins/PluginSearcher.scala
+++ b/plugins/src/main/scala/org/apache/toree/plugins/PluginSearcher.scala
@@ -30,8 +30,12 @@ class PluginSearcher {
   /** Represents logger used by plugin searcher. */
   private val logger = LoggerFactory.getLogger(this.getClass)
 
+  /** Contains all internal class information for matching plugins. */
+  private lazy val internalClassInfo: Map[String, ClassInfo] =
+    loadClassMap(newClassFinder())
+
   /** Contains all internal plugins for the system. */
-  lazy val internal: Seq[ClassInfo] = findPluginClasses(newClassFinder()).toSeq
+  lazy val internal: Seq[ClassInfo] = findPluginClasses(internalClassInfo).toSeq
 
   /**
    * Searches in the provided paths (jars/zips/directories) for plugin classes.
@@ -40,7 +44,7 @@ class PluginSearcher {
    * @return An iterator over plugin class information
    */
   def search(paths: File*): Iterator[ClassInfo] = {
-    findPluginClasses(newClassFinder(paths))
+    findPluginClasses(loadClassMap(newClassFinder(paths)))
   }
 
   /**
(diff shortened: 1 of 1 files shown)


danrope:
Thanks!  In addition, we've modified the Brunel magic to move out of the Toree namespace, updated to compile against latest Toree and confirmed this works.  We will release this change in Brunel 1.2.

lbustelo:
@chipsenkbeil I merged #39. You should be able to add the plugin test using the kernel tester.

chipsenkbeil:
@lbustelo, @poplav, I haven't looked at how the system tests are set up yet. How difficult is it going to be to get a jar in the system tests? I want to be able to add a magic using %AddJar and verify that it works and don't want anything related to the network to be involved.

poplav:
@chipsenkbeil , The system tests are [run](https://github.com/apache/incubator-toree/blob/master/Makefile#L181) in a docker container.  You should be able to volume mount some sort of test resources dir [here](https://github.com/apache/incubator-toree/blob/master/Makefile#L184) and operate normally.  You would put your test case as is [here](https://github.com/apache/incubator-toree/blob/master/test_toree.py#L41).  I assume this, but have not tested it yet.  
My main concern is as of now those tests  run code and expect an exact output.  In the case of AddJar unpredictable output might occur, but that is probably not the case because it would be local/not network.  Is there a way to suppress output for certain lines? If so that may help.

chipsenkbeil:
@poplav, can you provide code and ignore the result? I don't care about the output cell of the `%AddJar` call. I want to instead test the resulting side effect.

poplav:
@chipsenkbeil, Currently I don't think jupyter_kernel_test can achieve that.  I'm going to open an issue on jupyter_kernel_test which will use a custom asserter that can be passed in from the client.  If the user defined asserter is present you would use that over strict equality.  The assertion logic is [here](https://github.com/jupyter/jupyter_kernel_test/blob/b51d9911bc7a8bfbb826d3fd3149767c319b73fe/jupyter_kernel_test/__init__.py#L212).

So the proposed change would make it so you can do something like:
```
    code_execute_result = [
        {'code': '6*7', 'result': '42', 'custom_asserter': your_custom_asserter_defined function}
    ]
```

chipsenkbeil:
Sounds good to me. @lbustelo, since the jupyter tests aren't quite ready for me to add this specific test, can I go ahead and merge this?

lbustelo:
@chipsenkbeil Open a JIRA to add the system test and place it on the 0.1.0 release bucket.

lbustelo:
@danrope @chipsenkbeil 
Tested the PR. Looks good... see below

magic_plugin_works

Will merge and push a dev8 out.

PR #43, opened 2016-05-27
Add jar kernel test and override execute_helper

poplav:
[TOREE 317](https://issues.apache.org/jira/browse/TOREE-317?jql=project%20%3D%20TOREE)
* Overrides execute_helper to allow handling multiline code/streaming output that could exist in an add jar type operation.
* Includes sample custom test other than what is built it in.
* Tests addjar by volume mounting the resource/file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,9 +182,10 @@ system-test: pip-release
 		--name jupyter_kernel_tests \
 		-v `pwd`/dist/toree-pip:/srv/toree-pip \
 		-v `pwd`/test_toree.py:/srv/test_toree.py \
+		-v `pwd`/scala-interpreter/src/test/resources:/srv/system-test-resources \
 		$(IMAGE) \
-		bash -c "pip install /srv/toree-pip/toree*.tar.gz ; jupyter toree install --user --kernel_name='Apache_Toree' ; \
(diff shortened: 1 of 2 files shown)


poplav:
@lbustelo, [TOREE-318 (about the pyspark interpreter execute result)](https://issues.apache.org/jira/browse/TOREE-318) is almost done.  I am planning on adding kernel tests for that similar to what is here, thus want to make sure this looks right before continuing.

chipsenkbeil:
Deferring to @lbustelo after pointing out the local copy of `lwjgl-3.0.0b.jar`. Originally, we were verifying that downloading a jar worked along with adding it. Hosting it locally removes the network risk but also removes the test involving checking that downloading works.

So, do we want to a) test a local jar or a network jar and b) do we want to continue using lwjgl or switch to our own jar?

poplav:
Amended the commit to download the jar instead of using it locally.

poplav:
@chipsenkbeil Can you give this a final review and merge?  This is hosting system-test resources (`TestJar.jar`) in the system test container using a python http.server.  This was the jar we already were using for testing in `scala-interpreter/src/test/resources/`.

chipsenkbeil:
LGTM

PR #46, opened 2016-06-04
Moving from a Build.scala file to a build.sbt file.

fiskl:
I've trying to shift the build information to a build.sbt file, which is easier to read and understand.


fiskl:
Currently WIP, need to get the makefile to use the new build.sbt

chipsenkbeil:
@lbustelo, @Lull3rSkat3r, this is pretty nice considering sbt 1.0+ has removed support for `Build.scala` in favor of `build.sbt`. So, we'll need to do this eventually.

Lull3rSkat3r:
LGTM, the only feedback I have is about the `BuildInfo` class. It looks like the package name was changed. If we can we should not move the package as to not break any references people have to that class. Otherwise great job!

fiskl:
I've rewritten history and moved the BuildInfo package back to org.apache.toree.kernel.

chipsenkbeil:
LGTM

PR #47, opened 2016-06-06
Fix pyspark interpreter to handle execute results and output streams correctly.

poplav:
[TOREE-318](https://issues.apache.org/jira/browse/TOREE-318)
* Passes the output stream into the BrokerState and maps it to the code_id, so that the interpreter can call sendOutput with the code_id and write to the correct output stream.
* Makes use of python's AST parser to figure out if the last code statement in the request is an expression, if its an expression it will send output to the stream immediately from stdout, else assign the last statement to a temp variable where it is evaluated and if that has output return it as an execute result.
* Added system test coverage using the Jupyter kernel tester, which will test a print statement and a execute result producing arithmetic expression.  Note the kernel tester will test pyspark now as well as scala spark.

### DIFF
--- a/kernel-api/src/main/scala/org/apache/toree/interpreter/broker/BrokerState.scala
+++ b/kernel-api/src/main/scala/org/apache/toree/interpreter/broker/BrokerState.scala
@@ -22,7 +22,8 @@ import java.util.concurrent.ConcurrentHashMap
 import org.apache.toree.interpreter.broker.BrokerTypes._
 import org.slf4j.LoggerFactory
 
-import scala.concurrent.{Promise, Future, promise}
+import scala.concurrent.{Future, Promise}
+import scala.tools.nsc.interpreter.OutputStream
 
 /**
  * Represents the state structure of broker.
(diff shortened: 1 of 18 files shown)


kernel-api/src/main/scala/org/apache/toree/interpreter/broker/BrokerState.scala
```diff
@@ -69,10 +72,21 @@ class BrokerState(private val maxQueuedCode: Int) {
     // Add the code to be executed to our queue and the promise to our map
     codeQueue.add(brokerCode)
     promiseMap.put(brokerPromise.codeId, brokerPromise)
-
+    // Maintain an output stream for each codeId
+    outputResultStreamMap.put(brokerPromise.codeId, outputResultStream)
     codeExecutionPromise.future
   }
 
+  // Called from pyspark
```
chipsenkbeil: Can you provide a real Scaladoc for this versus "Called from pyspark"? Something more descriptive, but not specific to pyspark either since it's in the generic `BrokerState` class.

kernel-api/src/main/scala/org/apache/toree/interpreter/broker/BrokerState.scala
```diff
@@ -49,7 +52,7 @@ class BrokerState(private val maxQueuedCode: Int) {
    *
    * @return The future containing the results of the execution
    */
-  def pushCode(code: Code): Future[CodeResults] = synchronized {
+  def pushCode(code: Code, outputResultStream: Option[OutputStream]): Future[CodeResults] = synchronized {
```
chipsenkbeil: Rather than requiring None as a second argument everywhere, can you have this be an overloaded method where one instance has a signature of `def pushCode(code: Code)` with the output stream set to None or just have the `outputResultStream: Option[OutputStream]` be set to a default value of None in this function? Don't want to have to put None in explicitly everywhere.

poplav:
@chipsenkbeil , I think your comments have been addressed.  If it looks good can you merge.

chipsenkbeil:
LGTM

PR #48, opened 2016-06-06
Trivial: Removing redundant code from kernel bootstrap

fiskl:


chipsenkbeil:
Interesting. It's been awhile since the lazy Spark Context initialization support was added. Clearly, the shutdown doesn't report any problem with the Spark Context being null because it's wrapped in a `Try` that tosses any errors. The question is, "Do we want to remove the code or add back support for shutting down the Spark Context when the JVM is being shutdown?" Since it doesn't seem to be needed, I'm in favor of merging in this PR that just removes the failing code inside the shutdown hook.

@lbustelo, @Lull3rSkat3r, thoughts?

Lull3rSkat3r:
There is a discussion about if stop is import on the [Spark Mailing List](http://apache-spark-user-list.1001560.n3.nabble.com/SparkContext-stop-td17826.html). Matei explicitly says it is ok to not call stop. However, if the last statement is true we may want to add in the option to stop the context on shutdown.

As of now, I am in favor of merging. 👍 from me, thanks for the PR!

chipsenkbeil:
@fiskl, can you rebase this PR again? Want to keep the commit history relatively clean.

fiskl:
Rebased off the latest

chipsenkbeil:
LGTM

chipsenkbeil:
@fiskl, not sure if it matters to you, but the email address you're using for your commits doesn't seem to be associated with your Github account. Commits will go in fine, but Github won't associate them with your account unless you add the email address you used to your account.

fiskl:
That's a good point, git global settings + work email. Have to remember to change that, cheers

PR #50, opened 2016-06-11
Wrapping the kernel object in a python class

fiskl:
The current python runner is a touch broken, not in the least due to an oddity that can lead to multiple gateways if you create a Python SparkConf object.

This PR aims to address that issue, and as a useful consequence permit more pythonic use of the kernel object:
```
conf = (
  SparkConf()
            .setMaster("local[1]")
            .setAppName("Jupyter PySpark")
)
kernel.createSparkContext(conf)
```

### DIFF
--- a/pyspark-interpreter/src/main/resources/PySpark/pyspark_runner.py
+++ b/pyspark-interpreter/src/main/resources/PySpark/pyspark_runner.py
@@ -94,6 +94,13 @@ class Kernel(object):
     def __init__(self, jkernel):
         self._jvm_kernel = jkernel
 
+    def __getattr__(self, name):
+        return self._jvm_kernel.__getattribute__(name)
+
+    def __dir__(self):
+        parent = super().__dir__()
+        return parent + [x for x in self._jvm_kernel.__dir__() if x not in parent]
+
     def createSparkContext(self, config):
         jconf = gateway.jvm.org.apache.spark.SparkConf(False)
         for key,value in config.getAll():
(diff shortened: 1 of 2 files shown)


fiskl:
I expect this PR will lead to a bit of back-and-forth, as it is effectively creating a new API in the python kernel object. Thought it was worth starting that discussion as I think hiding the py4j bridge as much as possible is a worthy goal

chipsenkbeil:
Is there any way in Python to add a "catch all" method to a class such that invocations on that class that do not match methods like `refreshContext` will be captured? I'm assuming the wrapper kernel shadows the existing one. Ideally, when not providing a wrapper method, it would be nice to fall back to the underlying kernel instance.

fiskl:
Seems like overriding `__getattr__` and `__setattr__` might do what you want?

chipsenkbeil:
Does `__getattr__` get called during method invocation as well (as in, this would look up the method and return it to then be invoked). If so, then yeah, I would think that could do it.

fiskl:
Snippet showing the general attribute delegation pattern:
```
class Foo(object):
    def __init__(self):
        self._bar = "hey"
        
    def bar(self):
        return self._bar

class Bar:
    def __init__(self):
        self._foo = Foo()
        
    def baz(self):
    	return "hey to you too"
        
    def __dir__(self):
    	parent = super().__dir__()
    	
    	return parent + [x for x in self._foo.__dir__() if x not in parent]
    	
    def __getattr__( self, name):
    	return self._foo.__getattribute__(name)

bar = Bar()

print(bar.bar())
print(dir(bar))
```

chipsenkbeil:
@fiskl, is this finished?

fiskl:
Yes, it is

PR #51, opened 2016-06-18
Moving language info to the Interpreter object

fiskl:
Currently WIP, but because my Scala fu is weak I thought I'd get a consult.

The code here works, but has ~~two~~ one thing~~s~~ I want to fix before merging.
1. Python version is hard-coded. I wasn't sure where to get that information from?
2. Edit: I learned a thing, and stuck in a case class ~~The languageInfo object is a `Map`. I'd prefer to have some object (to allow for arbitrary nesting), but to be honest I've never serialized JSON in Scala so I don't know the best way to approach this.~~

What are your thoughts?

### DIFF
--- a/pyspark-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/pyspark/PySparkInterpreter.scala
+++ b/pyspark-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/pyspark/PySparkInterpreter.scala
@@ -89,8 +89,7 @@ class PySparkInterpreter(
    *         execution or the failure
    */
   override def interpret(code: String, silent: Boolean, output: Option[OutputStream]):
-    (Result, Either[ExecuteOutput, ExecuteFailure]) =
-  {
+    (Result, Either[ExecuteOutput, ExecuteFailure]) = {
     if (!pySparkService.isRunning) pySparkService.start()
 
     val futureResult = pySparkTransformer.transformToInterpreterResult(
@@ -155,11 +154,22 @@ class PySparkInterpreter(
   // Unsupported
   override def doQuietly[T](body: => T): T = ???
 
-  // TODO Identify how to plumb python version to here
-  override def languageInfo = LanguageInfo(
-    "python",
-    pySparkState.getVersion(),
-    fileExtension = Some(".py"),
-    pygmentsLexer = Some("ipython2"))
+  override def languageInfo: LanguageInfo = {
+    if (!pySparkService.isRunning) pySparkService.start()
 
+    import scala.util.control.Breaks._
+    val waitLimit = System.currentTimeMillis() + java.util.concurrent.TimeUnit.SECONDS.toMillis(5)
+    while (!pySparkState.isReady) {
+      if (System.currentTimeMillis > waitLimit) {
+        logger.warn("Timed out waiting for broker state to become ready")
+        break
+      }
+    }
+
+    LanguageInfo(
+        "python",
+        pySparkState.getVersion(),
+        fileExtension = Some(".py"),
+        pygmentsLexer = Some("ipython2"))
+  }
 }


dsaradini:
Hard to give you a thought without the big picture. 
If I understand well, toree is meant to execute remotely a snippet of code (in a specified language) and return back the result? 

I am certainly very extremely naive but, does the language not coming from the message sent through the IPython communication protocol? ( https://ipython.org/ipython-doc/3/development/messaging.html#kernel-info ) Certainly given by the Notebook's cell and/or, if we want to execute our own piece of code, the message we send to the toree server?

The fact that the language_info is a Map is certainly due to that it comes from a dynamic parsing of the JSON (IPython protocol). If you want to be strict you will have to add the 3 missing fields (mimetype, codemirror_mode, nbconvert_exporter,... ) and maybe to add a comment linking to the documentation in the code.

The real challenge here would be create a "simple" ReST interface in our server to communicate with a Toree server to execute piece of code from our Web UI rights?

chipsenkbeil:
@fiskl, is this ready to be merged? Haven't been keeping up with the conversation on this one. Not sure if you had addressed @dsaradini or tackled the version retrieval that you wanted to support on the broker. I see that you've put in version information for Scala and Python while leaving SparkR and Sql TBD, which is fine.

fiskl:
No - I'm happyish with the Interpreter api, but need to investigate further how to plumb the Python version up from the broker. Been too sick to tackle it this week

fiskl:
I am kinda tempted to simply add a method to the BrokerState that gets/sets properties, a little bit of shared memory between the parent and child processes. The pyspark_runner could then set a `languageVersion` property, and the PysparkInterpreter could use that in `languageInfo`.

I'm not across the jupyter interaction, will there be race conditions if this is done?

What do you two think?

chipsenkbeil:
@fiskl, we should be able to avoid race conditions, although it might take the kernel a little more time to respond if the language isn't baked into the interpreter as a constant.

While being able to get/set arbitrary properties could be useful, I think the language information should be a specific method (maybe just using property settings underneath) since that should be supported across all languages including Python and R.

fiskl:
First cut of this branch is ready to merge, if the build passes.

I'm simply removed version information for now, as it is not strictly required. The only required `language_info` field is `name`.

 When I communicated to version back from the broker state I ran into a race condition (or an invalid assumption). The data was requested before the broker state was set to ready, so the version never came back up.

I'll look further into it at leisure

fiskl:
It is a bit ugly, but you gotta do what you gotta do. In the PySpark interpreter I had to await the broker state being ready before returning the languageinfo object.

Also, treated `version` as special, forcing it to be returned when you mark the broker state as ready.

Finally, I notice that nobody ever actually calls the start method on the non-scala interpreters, it only does it internally

fiskl:
@chipsenkbeil Build passed, and this is now able to be merged

pyspark-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/pyspark/PySparkInterpreter.scala
```diff
@@ -155,11 +154,22 @@ class PySparkInterpreter(
...
-    "python",
-    pySparkState.getVersion(),
-    fileExtension = Some(".py"),
-    pygmentsLexer = Some("ipython2"))
+  override def languageInfo: LanguageInfo = {
+    if (!pySparkService.isRunning) pySparkService.start()
 
+    import scala.util.control.Breaks._
+    val waitLimit = System.currentTimeMillis() + java.util.concurrent.TimeUnit.SECONDS.toMillis(5)
+    while (!pySparkState.isReady) {
```
chipsenkbeil: Now that I see this, it's not good to have a loop like this (even if the max timeout is  5 milliseconds). So, two comments with this:

1. Move the `java.util.concurrent.TimeUnit.SECONDS.toMillis(5)` to a constant. Maybe just create a PySparkInterpreter companion object and put the constant in there.
2. Add a `Thread.sleep(1)` or something to your while loop at the end. This a) frees up the CPU to handle other tasks and b) enables the while loop to be interruptible.

If you add those two things, I suppose we can add this for now, but it does feel like a hack.

fiskl:
Is there a better way to do it? I feel like the start + stop methods should be explicitly called, rather than passively initializing the interpreter?

chipsenkbeil:
@fiskl, so I totally goofed and never replied to your comment. Would you still like to tackle this? If so, I'll take another look through your code to refresh myself.

sql-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/sql/SqlInterpreter.scala
```diff
@@ -112,4 +113,7 @@ class SqlInterpreter() extends Interpreter {
 
   // Unsupported
   override def doQuietly[T](body: => T): T = ???
+
+  override def languageInfo = LanguageInfo("scala", BuildInfo.scalaVersion, fileExtension = Some(".scala"))
```
mariusvniekerk: This should probably be 

```
LanguageInfo("sparkSql", BuildInfo.sparkVersion, fileExtension = Some(".sql"))
```

sparkr-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/sparkr/SparkRInterpreter.scala
```diff
@@ -144,4 +145,7 @@ class SparkRInterpreter(
 
   // Unsupported
   override def doQuietly[T](body: => T): T = ???
+
+  override def languageInfo = LanguageInfo("scala", BuildInfo.scalaVersion, fileExtension = Some(".scala"))
```
mariusvniekerk: Should this rather be

```scala
LanguageInfo("Rl", BuildInfo.sparkVersion, fileExtension = Some(".Rscript"), pygmentsLexer="R")
```

mariusvniekerk:
@fiskl @lbustelo @chipsenkbeil Do you guys want me to rebase this over to the master branch?

chipsenkbeil:
It can be rebased, but I don't want to merge it in since it has while loops checking a ready states. I think that needs to be reworked.

mariusvniekerk:
Yeah I'll clean it up a bit with the comments In here.

PR #57, opened 2016-06-23
Remove @Experimental from Comm classes

krishnakalyan3:
https://issues.apache.org/jira/browse/TOREE-263


chipsenkbeil:
@krishnakalyan3, can you continue to close and reopen until the tests pass? They ended up hanging again. You don't need to delete your branch. When you click close, you should have a button to reopen as well.

lbustelo:
I would leave the experimental on. Would be nice to make the Comm API more similar to the python api.

PR #60, opened 2016-09-07
Fix missing py4j_zip error 

siudzinskim:
Fixes missing variable declaration if py4j not available

### DIFF
--- a/etc/pip_install/toree/toreeapp.py
+++ b/etc/pip_install/toree/toreeapp.py
@@ -91,6 +91,7 @@ class ToreeInstall(InstallKernelSpec):
         try:
             py4j_zip = list(filter( lambda filename: "py4j" in filename, python_lib_contents))[0]
         except:
+            py4j_zip = None
             self.log.warn('Unable to find py4j, installing without PySpark support.')
 
         kernel_spec = KernelSpec()


etc/pip_install/toree/toreeapp.py
```diff
@@ -91,6 +91,7 @@ def create_kernel_json(self, location, interpreter):
         try:
             py4j_zip = list(filter( lambda filename: "py4j" in filename, python_lib_contents))[0]
         except:
+            py4j_zip = None
```
mariusvniekerk: So rather than just setting this to `None` here, can you change the kernel_spec.env lower down to not set PYTHON_PATH, since it will be incorrect in this case.

aquincum:
I'm being bitten by this very issue myself, a fix would be appreciated

n-marion:
Has this been added to a release yet?

devang:
On Ubuntu 18.04 I had to pass in a couple of flags, and everything worked fine after:

pip3 install --user py4j pyspark

jupyter toree install --spark_home=/home/<username>/.local/lib/python3.6/site-packages/pyspark --ToreeInstall.prefix=/home/<username>/.local

PR #62, opened 2016-09-13
[TOREE-343] Added support for `is_complete_request`

mariusvniekerk:
According to the spec here we need to implement support for is_complete_request

https://github.com/jupyter/jupyter_client/blob/4.4.0/docs/messaging.rst#code-completeness

This is i think the minimal amount required to support this.

### DIFF
--- a/kernel-api/src/main/scala/org/apache/toree/interpreter/Interpreter.scala
+++ b/kernel-api/src/main/scala/org/apache/toree/interpreter/Interpreter.scala
@@ -110,6 +110,13 @@ trait Interpreter {
    */
   def completion(code: String, pos: Int): (Int, List[String] )
 
+
+  /**
+    * Attempt to determine if a multiline block of code is complete
+    * @param code The code to determine for completeness
+    */
+  def isComplete(code: String): (String, String)
+
   /**
    * Returns the name of the variable created from the last execution.
    * @return Some String name if a variable was created, otherwise None
--- a/kernel-api/src/main/scala/org/apache/toree/interpreter/Interpreter.scala
+++ b/kernel-api/src/main/scala/org/apache/toree/interpreter/Interpreter.scala
@@ -23,6 +23,7 @@ import org.apache.toree.kernel.api.KernelLike
 
 import scala.tools.nsc.interpreter._
 
+
 trait Interpreter {
 
   /**
@@ -108,14 +109,13 @@ trait Interpreter {
    * @param pos The cursor position
    * @return The cursor position and list of possible completions
    */
-  def completion(code: String, pos: Int): (Int, List[String] )
-
+  def completion(code: String, pos: Int): (Int, List[String] ) = (pos, Nil)
 
   /**
     * Attempt to determine if a multiline block of code is complete
     * @param code The code to determine for completeness
     */
-  def isComplete(code: String): (String, String)
+  def isComplete(code: String): (String, String) = ("unknown", "")
 
   /**
    * Returns the name of the variable created from the last execution.
--- a/kernel/src/main/scala/org/apache/toree/kernel/protocol/v5/interpreter/InterpreterActor.scala
+++ b/kernel/src/main/scala/org/apache/toree/kernel/protocol/v5/interpreter/InterpreterActor.scala
@@ -58,6 +58,7 @@ class InterpreterActor(
   //
   private var executeRequestTask: ActorRef = _
   private var completeCodeTask: ActorRef = _
+  private var isCompleteTask: ActorRef = _
 
   /**
    * Initializes all child actors performing tasks for the interpreter.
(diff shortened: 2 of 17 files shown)


lbustelo:
@chipsenkbeil Take a look at this pr. Would be nice if we can cherry-pick parts of it into the 0.1.0 branch.

chipsenkbeil:
Looking at it now.

mariusvniekerk:
Need to still fix it up a bit

chipsenkbeil:
Looks like you got a passing build. Ready for me to take a look?

mariusvniekerk:
Yep.  Go for it.

kernel/src/main/scala/org/apache/toree/kernel/protocol/v5/interpreter/InterpreterActor.scala
```diff
@@ -96,5 +99,13 @@ class InterpreterActor(
             ex.getStackTrace.map(_.toString).toList)
           )
       } pipeTo sender
+    case (isCompleteRequest: IsCompleteRequest) =>
+      logger.debug(s"InterpreterActor requesting is complete code ${isCompleteRequest.code}")
+      (isCompleteTask ? isCompleteRequest) recover {
+        case ex: Throwable =>
+          logger.warn(s"Could determine completeness for code ${isCompleteRequest.code}: " +
```
chipsenkbeil: Shouldn't this say "Could not determine"?

mariusvniekerk: yes

chipsenkbeil:
@lbustelo, other than the bit about the 2.10 vs 2.11 changes, the rest of this could be put into the 1.x branch, but I'm not sure how hard that is to do with Git when certain files don't exist. For the 1.x branch, the kernel would always say "unknown" for the status, since only 2.11 has isComplete fully implemented.

kernel/src/main/scala/org/apache/toree/kernel/protocol/v5/interpreter/InterpreterActor.scala
mariusvniekerk: Fixed, not sure why github is still showing the outdated diff

chipsenkbeil: Looks like each interpreter should at least return an unknown status, so I think a warn is appropriate to bring to the surface that something is going wrong with the check. Furthermore, we log an error for the other message routes, so I think a warn is appropriate considering we are not bubbling up an exception.

kernel-api/src/main/scala/org/apache/toree/interpreter/Interpreter.scala
```diff
@@ -110,6 +110,13 @@ trait Interpreter {
    */
   def completion(code: String, pos: Int): (Int, List[String] )
 
+
+  /**
+    * Attempt to determine if a multiline block of code is complete
+    * @param code The code to determine for completeness
+    */
+  def isComplete(code: String): (String, String)
```
mariusvniekerk: Since every kernel should return at least the unknown status, should we do that here instead?

chipsenkbeil: Makes sense to me. Maybe we should make a companion object that contains the default isComplete. Something like...

```
object Interpreter {
  val DefaultIsCompleteResult: (String, String) = ("unknown", "")
}

trait Interpreter {
  // ...
  def isComplete(code: String): (String, String) = Interpreter.DefaultIsCompleteResult
  // ...
}
```

mariusvniekerk: We could just define a default concrete implementation in the trait and not need the companion class, your call.

chipsenkbeil: I'm fine either way. Just go with your own preference.

mariusvniekerk: I'll pick the option that lets me nuke more lines of code 💥

mariusvniekerk:
@chipsenkbeil Review phase 2

chipsenkbeil:
LGTM. Can you rebase this one on top of master as well? I'll merge once you have done that.

chipsenkbeil:
Is this a hiccup in our tests? Or did something actually fail?

```
======================================================================

ERROR: test_scala_execute_result (__main__.ToreeScalaKernelTests) (code='6*7')

Asserts test_statements execute correctly meaning the last message is the expected result

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/srv/test_toree.py", line 87, in test_scala_execute_result

    self.assertIn('text/plain', output_msgs[-1]['content']['data'])

IndexError: list index out of range

======================================================================

ERROR: test_scala_execute_result (__main__.ToreeScalaKernelTests) (code='%showtypes on\n1')

Asserts test_statements execute correctly meaning the last message is the expected result

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/srv/test_toree.py", line 87, in test_scala_execute_result

    self.assertIn('text/plain', output_msgs[-1]['content']['data'])

KeyError: 'data'
```

chipsenkbeil:
This will need another rebase once it's ready (not sure if it is based on the failing tests).

mariusvniekerk:
Yeah, I think there is some order-dependant pieces in jupyter kernel test.

Rebased

PR #63, opened 2016-09-21
TOREE-341 Fixed exception stack traces getting lost.

mariusvniekerk:
This is a rather ugly fix to address the issue caused by
https://issues.scala-lang.org/browse/SI-8935

The crux of the issue is that valueOfTerm returns None -- always

This introduces a small class that is bound into the REPL scope so that we
can modify its internal state and use that to retrieve values.

We can extend this to a more general case later on if needed.

### DIFF
--- a/kernel/src/test/scala/integration/PostProcessorSpecForIntegration.scala
+++ b/kernel/src/test/scala/integration/PostProcessorSpecForIntegration.scala
@@ -19,10 +19,10 @@ package integration
 
 import java.io.OutputStream
 
-import org.apache.toree.kernel.api.KernelLike
+import org.apache.toree.kernel.api.Kernel
 import org.apache.toree.kernel.interpreter.scala.ScalaInterpreter
 import org.apache.toree.kernel.protocol.v5.magic.PostProcessor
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
This seems to be causing sporadic test failures with jupyter_kernel_test.  When I run it locally they pass sometimes.  I suspect there is some kind of ordering/timing issue with how results are sent back to the client?

mariusvniekerk:
Looks like @transient did the trick? maybe?

mariusvniekerk:
@chipsenkbeil @lbustelo Ready for review.

chipsenkbeil:
Quick glance seems fine. Shame about the hack since it introduces more code compilation and execution (and as we know the Scala REPL doesn't unload memory or classes), but it seems necessary because of the bug. @lbustelo, you fine with me merging this?

lbustelo:
Merge it

Gino B.

chipsenkbeil:
@mariusvniekerk, can you rebase on top of master for this one? Once you do, I'll merge.

mariusvniekerk:
Rebased

PR #64, opened 2016-09-22
The default spark application name for Toree is still IBM Spark Kernel.

mariusvniekerk:
Lets fix that.


mariusvniekerk:
I can cherry pick this for 0.1x if we want.

mariusvniekerk:
@chipsenkbeil Ready for review since this is just a one-line string change.

chipsenkbeil:
Ha ha, I had forgotten about this. Good catch. Yeah, we should backport this into the 0.1 branch as well. Don't want a release without the proper name.

mariusvniekerk:
I see that is also done in #51, but not sure on the status of that PR

chipsenkbeil:
A one-liner will be simple enough to rebase on top of, so I'm not worried. That reminds me that I was supposed to follow up on that.

PR #66, opened 2016-09-23
[TOREE-342] Switch completer

mariusvniekerk:
Switching the scala completer that we use to PresentationCompilerCompleter (which is the default one in Scala 2.11).  This should allow us to get much better autocompletion.


mariusvniekerk:
Test failure seems to be a red herring?  unrelated.

chipsenkbeil:
LGTM

PR #71, opened 2016-09-30
Feature/build fixes

petrosrizos:
Some small fixes for your Makefile and Docker build options: 
-ti -it flags are not needed, just -t for attached, otherwise you are binding your self to needing an TTY terminal which breaks in build servers such as Jenkins and other integration environments.

gpg should not be hard coded, but should be allowed to take advantage of the path, otherwise it causes portability issues for the build.


chipsenkbeil:
LGTM. Thanks!

PR #72, opened 2016-10-06
Fix for cell and line magics

bpburns:
Magics are not working properly on the scala 2.11 / spark 2.0 master branch. For example using the html magic:

`%%html`
`<b>foo</b>`

will return 

`Left(Map(text/html -> <b>foo</b>))`

The problem is in the scala IMain repl used in the 2.11 version of ScalaInterpreterSpecific read method. IMain valueOfTerm is not working properly when given a variable name. IMain eval is working however.

### DIFF
--- a/scala-interpreter/src/main/scala-2.11/org/apache/toree/kernel/interpreter/scala/ScalaInterpreterSpecific.scala
+++ b/scala-interpreter/src/main/scala-2.11/org/apache/toree/kernel/interpreter/scala/ScalaInterpreterSpecific.scala
@@ -258,12 +258,11 @@ trait ScalaInterpreterSpecific extends SettingsProducerLike { this: ScalaInterpr
    */
   override def read(variableName: String): Option[AnyRef] = {
     require(iMain != null)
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
The other option you can try is similar to how we capture the exceptions.  It's not pretty but sadly due to some bugs in the Scala Interpreter unavoidable.  See the ExceptionHack class in that interpreter.

mariusvniekerk:
Not sure if the switch from types of AnyRef to String is going to cause us future hassles.

bpburns:
IMain eval returns an AnyRef.  What return String were you referring to?

mariusvniekerk:
ah, think the special String handling in the pattern match threw me.

bpburns:
Yeah, that just making sure that if it's an empty String return None.  Most of the type the return values ends up being Either[CellMagic, LineMagic] or whatever those class names are.

chipsenkbeil:
LGTM. Need to get this in pronto, so I'll look at adding tests to verify this automatically later.

PR #75, opened 2016-10-17
[TOREE 350] Changes to address comments on rc2

lbustelo:
- [x] no KEYS file so no way to verify the gpg signature ( unless one downloads an unverified key from a public server )
- [x] bin tarball should untar into an apache-toree-incubating or apache-toree directory, source tarball untars into ./ - should be fixed to use a top-level dir IMO.
- [ ] Something to check on - are the copyrights needed in the NOTICE file or the LICENSE file? I am not too sure if there are needed in the NOTICE file.
- [x] licenses/LICENSE-jline.txt seems to have some html but not the actual license content. Did not look at all the files so folks should re-check those.
- [ ] Most projects tend to have one license file per license type and not a license file per dependency - with the copyrights called out in the main LICENSE file I believe.
- [x] source tarball seems to have too many licenses. Unless jline/scala,asm.ammonite, etc are bundled into the source tarball, they do not need to be called out in the LICENSE and/or NOTICE file.
- [ ] bunch of markdown files without a license header
- [ ] there are a bunch of test jars checked into the source. Is there ALv2 provenance for all of them (including the sparkr tarball )?


mariusvniekerk:
So the sparkr tarball is removed in 0.2.x.

lbustelo:
Instructions of how to get the KEY at https://dist.apache.org/repos/dist/dev/incubator/impala/KEYS

lbustelo:
I do not see where other projects are adding a KEYS file in their source, so I'm only placing this on the dist.

chipsenkbeil:
@lbustelo, is this still relevant?

lbustelo:
@chipsenkbeil This is such a long time ago. I think there are some changes here that should make it to both the 0.1.x and master branch. More specifically:

1. The changes to remove legal files from src distro 
2. The changes to rename the package to `apache-toree`

lbustelo:
@chipsenkbeil rebased and removed WIP

chipsenkbeil:
LGTM

PR #77, opened 2016-10-19
Fix missing "test" modifier

okoethibm:
Missing "test" modifier causes jackson 2.7.4 libs to be unnecessarily included in the toree assembly.

If the toree assembly gets on the main Spark classpath (via spark.driver.extraClassPath)  this causes
       com.fasterxml.jackson.databind.JsonMappingException: Incompatible Jackson version: 2.7.4
...     at org.apache.spark.util.JsonProtocol$.<clinit>
because Spark 2.0 is built against a lower jackson version (2.6.5)

### DIFF
--- a/project/common.scala
+++ b/project/common.scala
@@ -146,7 +146,9 @@ object Common {
     // Test dependencies
     libraryDependencies ++= Seq(
       "org.scalatest" %% "scalatest" % "2.2.6" % "test", // Apache v2
-      "org.mockito" % "mockito-all" % "1.10.19" % "test"   // MIT
+      "org.mockito" % "mockito-all" % "1.10.19" % "test",   // MIT
+      // use the same jackson version in test than the one provided at runtime by Spark 2.0.0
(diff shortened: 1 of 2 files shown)


lbustelo:
Test failed with 

```
java.lang.NoClassDefFoundError: com/fasterxml/jackson/databind/Module
	at play.api.libs.json.Json$.stringify(Json.scala:51)
	at 

```

Looks suspicious.

lbustelo:
I think the fix is to add the jackson test dependency to https://github.com/apache/incubator-toree/blob/87a9eb8ad08406ce0747e92f7714d4eb54153293/communication/build.sbt

okoethibm:
adding to communication alone wan't sufficient. added the jackson test dep to base ToreeProject instead.
I still get some errors on "make test" but those are the same that I get in a baseline without my change

lbustelo:
It looks like travis is now happy.

lbustelo:
@okoethibm This had been merged both in `master` and in `0.1.x` branch. Can you close this PR?

PR #79, opened 2016-11-15
Updated to Coursier 1.0.0-M14

jpolchlo:
Changed references for Coursier 1.0.0-M12 to version 1.0.0-M14.  Required fixes due to API changes were also made.

### DIFF
--- a/kernel-api/src/main/scala/org/apache/toree/dependencies/CoursierDependencyDownloader.scala
+++ b/kernel-api/src/main/scala/org/apache/toree/dependencies/CoursierDependencyDownloader.scala
@@ -161,8 +161,8 @@ class CoursierDependencyDownloader extends DependencyDownloader {
    */
   override def removeMavenRepository(url: URL): Unit = {
     repositories = repositories.filterNot {
-      case MavenRepository(urlString, _, _, _, _) => url.toString == urlString
-      case _                                => false
+      case maven: MavenRepository => url.toString == maven.root
+      case _                      => false
     }
   }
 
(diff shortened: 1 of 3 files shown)


kernel-api/src/main/scala/org/apache/toree/dependencies/CoursierDependencyDownloader.scala
```diff
@@ -313,19 +313,19 @@ class CoursierDependencyDownloader extends DependencyDownloader {
    * @return The resulting URIs
    */
   private def repositoriesToURIs(repositories: Seq[Repository]) = repositories.map {
-    case IvyRepository(pattern, _, _, _, _, _, _, _, _)  => pattern
-    case MavenRepository(root, _, _, _)                  => root
+    case IvyRepository(pattern, _, _, _, _, _, _, _)  => pattern.string
```
mariusvniekerk: So i think that rather than having this pattern matching with all the underscores, we probably want to just do something like

```scala
case ivy: IvyRepository => ivy.pattern.string
```

and similar in other places.  That should allow a bit more robustness for when coursier changes again.

chipsenkbeil:
@mariusvniekerk, any reason why this hasn't been merged in?

mariusvniekerk:
This can probably be merged in.  We might want to bump slightly further since I think coursier is a few milestones ahead

jodersky:
let's do that. The latest version is 1.0.0-M15-1. Should we just merge this and update the M14 references with a rebase?

mariusvniekerk:
Sounds reasonable to me

jodersky:
Go ahead. I'll have some time for this tomorrow in case you're busy.

mariusvniekerk:
I'm finishing up prep for my talk on Toree tomorrow

PR #80, opened 2016-11-16
Issue 351 - delay calls to scala interpreter until all other toree in…

pberkland:
…itialization complete , prevents problems with contextClassLoader

### DIFF
--- a/kernel-api/src/main/scala/org/apache/toree/interpreter/Interpreter.scala
+++ b/kernel-api/src/main/scala/org/apache/toree/interpreter/Interpreter.scala
@@ -140,4 +140,9 @@ trait Interpreter {
     */
   def languageInfo: LanguageInfo
 
+  /**
+   * Initialization done after all other Toree initialization done.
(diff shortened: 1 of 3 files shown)


mariusvniekerk:
Thanks a lot for this one.

Is there some way we can reliably test this?

Probably the easiest spot is in the kernel gateway tests.

mariusvniekerk:
@chipsenkbeil @lbustelo Any objections to merging this guy.  Tested it locally and all seems well.

chipsenkbeil:
While I can't say I'm thrilled with adding another method to the interpreter trait, I'm not going to hold this one up.

Ideally, we would convert all of our startup code to use the plugin architecture. We'd then tack on some plugins to run when the SparkSession and SparkContext are available, which would then look up the interpreters and bind appropriately.

For now, this is okay. I'd say we add `@experimental` to the `postInit` method or mark in the docs that it could be dropped in later versions.

PR #82, opened 2017-01-04
Add Flush Method to Logger so it can be compliant to stream interface: JIRA TOREE-356 and TOREE-357

DTAIEB:
Logger doesn't have the flush method which is causing an exception when I try to use it as a stream for my project.

### DIFF
--- a/pyspark-interpreter/src/main/resources/PySpark/pyspark_runner.py
+++ b/pyspark-interpreter/src/main/resources/PySpark/pyspark_runner.py
@@ -151,12 +151,14 @@ kernel = Kernel(bridge.kernel())
 
 while True:
     try:
-        code_info = state.nextCode()
+        next_code_info = state.nextCode()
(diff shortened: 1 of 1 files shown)


DTAIEB:
There are 2 minor code changes in this PR:
-Add Flush Method to Logger object so it can be compliant with stream interface
-In the loop that pull code_info from the queue, make sure we have a new code_info before setting the current one to null. Reason is that in some cases (say Spark Streaming for example), the cell may have finished executing but there is still data coming down

mariusvniekerk:
Looks good to me

mariusvniekerk:
Can you make an issue on the Toree JIRA for this and link it in your PR title?

lbustelo:
@mariusvniekerk If this looks good to you. I think you should do the merge.

rgbkrk:
:shipit:

PR #83, opened 2017-01-11
Some0.1.x changes to docs and make

lbustelo:
This is some minor changes to fix:

1. pip command to install from apache dist server
2. removed adding legal docs to src dist package


chipsenkbeil:
LGTM

PR #86, opened 2017-01-17
[TOREE-360] fire allInterpretersReady and preRunCell events

DTAIEB:
allInterpretersReady event: fired when all interpreters have completed
their initialization
preRunCell event: fired every time a new cell is being run, unless the
silent flag is set

### DIFF
--- a/kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
+++ b/kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
@@ -88,6 +88,8 @@ trait StandardComponentInitialization extends ComponentInitialization {
     initializeSparkContext(config, kernel, appName)
 
     interpreterManager.initializeInterpreters(kernel)
+    
+    pluginManager.fireEvent("allInterpretersReady")
 
     val responseMap = initializeResponseMap()
 
--- a/kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
+++ b/kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
@@ -21,7 +21,6 @@ import java.io.File
 import java.net.URL
 import java.nio.file.{Files, Paths}
 import java.util.concurrent.ConcurrentHashMap
-
 import akka.actor.ActorRef
 import com.typesafe.config.Config
 import org.apache.spark.SparkConf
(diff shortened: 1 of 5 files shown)


DTAIEB:
I am currently writing a toree plugin that leverage the PySpark Interpreter to run the Pixiedust Python library auto-viz capability (https://github.com/ibm-cds-labs/pixiedust) from a Scala notebook. What's interesting about this feature is that it brings viz rendering using Matplotlib, seaborn, bokeh and others to scala users.

This PR is firing 2 events that are needed by my plugin (see comment above), with no side effect if no plugin is currently handling them.

mariusvniekerk:
Probably best to have a jira from a tracking perspective. 

@lbustelo @chipsenkbeil does it make sense to also add a post cell run event for the plug-in system?

chipsenkbeil:
Sure, why not. The plugin system was meant to accrue more event types to support varying cases. It seems like having a post run cell event to compliment the pre run cell event would make sense.

I'm starting to think we should have a common object to store the event names to make it less likely to mis-type the event. Something like...

```scala
object PluginEvents {
  /** Signalled when all interpreters have been initialized. */
  val AllInterpretersReady: String = "allInterpretersReady"

  /** Signalled before executing a block of code. */
  val PreRunCell: String = "preRunCell"

  /** Signalled after executing a block of code. */
  val PostRunCell: String = "postRunCell"
}
```

mariusvniekerk:
Yeah that seems like a good idea.  We might also want that same place to hold some notion of the type contract that implementers should expect?

mariusvniekerk:
@DTAIEB I believe that the scalatest for this would need to be adjusted since you have more messages floating around now.

DTAIEB:
@mariusvniekerk : I will look at updating the scalatests and create a JIRA to track.
@chipsenkbeil : I agree with the idea of centralizing the event names and we should include the existing events as well. I can look at making these changes if everyone agrees.

chipsenkbeil:
@mariusvniekerk, what did you have in mind for contracts?

@DTAIEB, go for it.

DTAIEB:
@mariusvniekerk It looks like the pluginsManager is not mocked in the scalatest causing a NullPointer exception. I fixed it by adding a check, probably best to have someone update the code to mock the pluginsManager.
Waiting for Travis to complete the tests (for some reason, the build has been stuck in Created mode for the last hour...)

DTAIEB:
Tests are passing

chipsenkbeil:
@mariusvniekerk, were you wanting @DTAIEB to add the `postRunCell` trigger? Or leave that for a separate PR? He's got the case object defined, but I don't see it being triggered anywhere.

DTAIEB:
@chipsenkbeil yes, I would prefer to add it in a separate PR.

chipsenkbeil:
Going ahead and merging this one in.

PR #88, opened 2017-01-23
[WIP] Maint/jarremoval

mariusvniekerk:
cc @chipsenkbeil @lbustelo 

This is a potential way to get rid of the compiled jar pieces on 0.1.x.

Let me know if you think this works for us.  Basically the same approach that apache spark uses.


chipsenkbeil:
LGTM

PR #90, opened 2017-01-30
[TOREE-365] Show REPL declarations in notebooks

jodersky:
Always print the result of an evaluation, regardless if is an expression that is printed as `resX: <expr>`, or a declaration that does not contain the "resX:" part, such as "defined class Foo".

### DIFF
--- a/scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
+++ b/scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
@@ -162,28 +162,23 @@ class ScalaInterpreter(private val config:Config = ConfigFactory.load) extends I
       interpretRec(code.trim.split("\n").toList, false, starting)
    }
 
-   def truncateResult(result:String, showType:Boolean =false, noTruncate: Boolean = false): String = {
-     val resultRX="""(?s)(res\d+):\s+(.+)\s+=\s+(.*)""".r
+   def truncateResult(result: String, showType: Boolean = false, noTruncate: Boolean = false): String = {
+     val resultRX = """(?s)(res\d+):\s+(.+)\s+=\s+(.*)""".r
 
      result match {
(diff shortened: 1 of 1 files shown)


scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
```diff
@@ -162,28 +162,23 @@ class ScalaInterpreter(private val config:Config = ConfigFactory.load) extends I
...
+           returnStr = r.getOrElse("").toString
+         }
+         if (showType) {
+           returnStr = s"$varType = $returnStr"
+         }
          returnStr
 
-       }
-       case _ => ""
+       case str => str
```
jodersky: this is the only functional change, the others are just styling

mariusvniekerk:
So I'm not entirely certain on this one.  

On the one hand, this more closely maps to what the Scala console does.  
The existing solution, however, more closely maps to the semantics of IPython.

I guess the real question is what the expectations of users are for this.  Most of the other heavily used jupyter kernels (ipykernel, irkernel, ijulia) will only return one output from a cell instead of a bunch.  

cc @lbustelo and @chipsenkbeil

mariusvniekerk:
Might be a better idea to have this in a magic similar to what `%showtypes on/off` does atm.

chipsenkbeil:
This should either be a CLI option, a togglable magic as @mariusvniekerk pointed out, or its own magic that evaluates code. We shouldn't change the default behavior since we are following the Jupyter protocol. Furthermore, changing such a fundamental action's response could break any consumers of Toree relying on the existing response handling.

If we add a toggleable option, the default should be as it currently is.

Lull3rSkat3r:
I agree with making this a toggleable option. This behavior is meant to
reflect that of other kernels and is an experience that users of kernels
expect.

jodersky:
The reason for this change was originally a quick fix to ease working with Toree Client. Since the latter is being discussed on the mailing list, I'll close this PR for now.

PR #95, opened 2017-02-08
Fix system tests

jodersky:
CI is currently failing due to system tests.

Check http://unix.stackexchange.com/questions/342403/openjdk-8-jre-headless-depends-ca-certificates-java-but-it-is-not-going-to-be for an explanation of the fix.


mariusvniekerk:
LGTM?

PR #96, opened 2017-02-13
TOREE-376: Add implicits to the Scala interpreter.

rdblue:
This defines Scala implicits commonly used with Spark SQL. This includes dataframe functions, like `map` or `sum`, common conversions like `Seq#toDF`, and the `$` prefix to coerce strings to column names.

### DIFF
--- a/scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
+++ b/scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
@@ -90,10 +90,11 @@ class ScalaInterpreter(private val config:Config = ConfigFactory.load) extends I
      this
    }
 
-   override def postInit(): Unit = {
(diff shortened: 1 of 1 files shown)


rdblue:
It looks like the build is failing, but not from this patch. The scala tests are all passing. I'm also happy to add a test for this, if someone can tell me where it should go.

chipsenkbeil:
@mariusvniekerk, does this look good to you? If so, @rdblue, can you rebase this?

rdblue:
Rebased

mariusvniekerk:
Okay looks good to me.  I think we did this initially to limit the size of the classes that got generated.  However the quality of life improvement that this gives is worth it IMO.

chipsenkbeil:
There may be other tricks we can do to reduce the generated classes as well, so I'll merge this in. Thanks @rdblue and @mariusvniekerk!

rdblue:
Thank you @chipsenkbeil and @mariusvniekerk for reviewing!

PR #99, opened 2017-02-13
TOREE-379: Fix code completion reply.

rdblue:
The start and end positions were swapped, causing the front-ends to
replace no characters from the partial symbol that is completed.


mariusvniekerk:
Hi, we've fixed some of the issues in our docker container running tests.  Can you close and reopen so that travis can run?

mariusvniekerk:
Looks good to me.  All relevant tests are passing.  The failure is due to a missing apache header in the gitattributes file so nothing major.

PR #100, opened 2017-02-14
TOREE-363: Fix Scala syntax highlighting.

rdblue:
This fixes Scala syntax highlighting in Jupyter notebooks. The kernel info reply needs to include a value for "codemirror_mode" that is recognized by CodeMirror. From looking at `CodeMirror.mimeModes` in the console, the correct one for Scala is `text/x-scala`. This also adds the mime type and uses the same value.


mariusvniekerk:
Thanks for this one

mariusvniekerk:
Can you make the codemirror changes to the other language info things in the other interpreters.

rdblue:
Thanks for merging this, @mariusvniekerk!

PR #104, opened 2017-02-16
TOREE-380: Allow interpreters to format output.

rdblue:
This branch has the changes that I made to our Toree distribution to allow interpreters to format output. The goal is to enable better integration when libraries are used with Toree, similar to conventions in IPython like `_repr_html_` methods.

In IPython, classes can define a `_repr_html_` method that is used to produce a `text/html` representation of an output object that is shown in the Jupyter notebook. This branch adds a `Displayer` class that can be registered with the Scala interpreter and updates the interpreter so that it fetches the last object of an interpreted block and calls an appropriate displayer to get its output representations. I think this is a reasonable way to maintain type safety and allow libraries to produce HTML outputs in notebooks. I'm also interested in using the plugin system to register these.

In addition, there are a few example displayer implementations:
* A SparkSession/SparkContext displayer that produces HTML links to the a Spark application
* A [Vegas](https://github.com/vegas-viz/Vegas) displayer for graphs
* A MagicOutput displayer to avoid the need for post-processing
* A default displayer for AnyRef that calls toString and (as proof-of-concept) uses reflection to call toHtml

Implementation required changing the way outputs from interpreters and blocks are handled. Rather than an interpreter returning text, it now returns a mime-type to content map (which has various type names). It also changes magic to produce MagicOutput, instead of requiring post-processing.


scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
```diff
@@ -18,37 +18,45 @@
...
 import scala.concurrent.{Await, Future}
 import scala.language.reflectiveCalls
 import scala.tools.nsc.Settings
 import scala.tools.nsc.interpreter.{IR, OutputStream}
 import scala.tools.nsc.util.ClassPath
-import scala.util.{Try => UtilTry}
+
+import org.apache.toree.kernel.protocol.v5.MIMEType
+import org.apache.toree.magic.MagicOutput
+import vegas.DSL.ExtendedUnitSpecBuilder
```
mariusvniekerk: Not sure if we want these here.

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
```diff
@@ -328,6 +466,10 @@ class ScalaInterpreter(private val config:Config = ConfigFactory.load) extends I
 
 object ScalaInterpreter {
 
+  val NamedResult = """(\w+):\s+([^=]+)\s+=\s*(.*)""".r
```
jodersky: Will this work when the result type contains an `=`, e.g. `res0: Int => Int = <function1>`?
Using a non-greedy match on anything preceding seems to work quite well `"""(?s)(res\d+):\s+(.+?)\s+=\s+(.*)""".r`

rdblue: No, this won't work with function types since it assumes that types are any series of characters before the =. In the short term, we may want to have a simpler match that just grabs the variable name. Since we confirm that it is a variable by looking it up in the interpreter anyway, it is probably easier to get the variable and then inspect it to get the information that could be parsed here.

Long term, I don't think that parsing a stream is the way to go but I don't see a much better way with the current scala REPL API.

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
rdblue: Agreed. I don't think we want to add a dependency on Vegas, but I wanted to include another example (one that was a requirement from our users) to motivate the displayer idea.

I think to be included in Toree, this would need some changes. Displayer probably doesn't need to be a class, but a function `T => Map[String, String]`. I'd also like to make registration play nicely with plugins, so you can add a library and it registers functions to display its objects.

rdblue:
I've extracted the displayer registration from this PR and created an independent project for it: https://github.com/rdblue/jupyter-repr-api

The idea is to make it so that libraries have a standard way to provide code to display objects, without relying on a specific kernel implementation.

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
```diff
@@ -18,37 +18,45 @@
...
+   private var kernel: KernelLike = _
+
    protected val logger = LoggerFactory.getLogger(this.getClass.getName)
 
    protected val _thisClassloader = this.getClass.getClassLoader
 
    protected val lastResultOut = new ByteArrayOutputStream()
 
-   protected val multiOutputStream = MultiOutputStream(List(Console.out, lastResultOut))
+   protected val multiOutputStream = lastResultOut
```
kevin-bates: At a minimum, this needs to remain a MultiOutputStream (of one item) else it will produce a compilation error in [here](https://github.com/apache/incubator-toree/blob/master/scala-interpreter/src/test/scala/integration/interpreter/scala/AddExternalJarMagicSpecForIntegration.scala#L41).  Removal of Console.out is further discussed in #116.

rdblue:
@mariusvniekerk, @jodersky, @kevin-bates, I've cleaned up this PR and I think it is ready to start reviewing.

The major changes are:

1. Interpreters output Map[String, String] so they can send more than just text.
2. Magics also produce Map[String, String]
3. There is no longer a need for a magic PostProcessor, so it has been removed
4. This uses `jupyter.Displayers` to show Scala objects and registers some methods for Spark

More about Jupyter displayers here: https://github.com/jupyter/jvm-repr

rdblue:
Good idea. I'll add some tests for the DF output functions, which produce both HTML and plain text.

rdblue:
@mariusvniekerk, I added a few tests in JVMReprSpec and got the rest of them working. Thanks for taking a look at this!

scala-interpreter/build.sbt
```diff
@@ -18,3 +18,4 @@ import sbt.Tests.{Group, SubProcess}
  */
 
 libraryDependencies ++= Dependencies.sparkAll.value
+libraryDependencies += "com.github.jupyter" % "jvm-repr" % "0.1.0"
```
chipsenkbeil: What does this do? Looking it up, I see that @rdblue was the author, but now it's under the jupyter umbrella. BSD-3 clause should be fine.

Is there a reason it's just used here and not all interpreters?

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaDisplayers.scala
```diff
@@ -0,0 +1,207 @@
...
+import org.apache.spark.sql.SparkSession
+import jupyter.Displayer
+import jupyter.Displayers
+import jupyter.MIMETypes
+import org.apache.toree.kernel.protocol.v5.MIMEType
+import org.apache.toree.magic.MagicOutput
+
+object ScalaDisplayers {
+
+  def ensureLoaded(): Unit = ()
```
chipsenkbeil: This does nothing? Why is this method here? Should we have a common interface that the `ScalaDisplayers` object inherits?

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
```diff
@@ -18,30 +18,34 @@
...
 import scala.tools.nsc.Settings
 import scala.tools.nsc.interpreter.{IR, OutputStream}
 import scala.tools.nsc.util.ClassPath
-import scala.util.{Try => UtilTry}
+import scala.util.matching.Regex
 
 class ScalaInterpreter(private val config:Config = ConfigFactory.load) extends Interpreter with ScalaInterpreterSpecific {
+  import ScalaInterpreter._
+
+  ScalaDisplayers.ensureLoaded()
```
chipsenkbeil: Why are we calling this if it does nothing?

kernel/src/test/scala/org/apache/toree/kernel/protocol/v5/stream/KernelInputStreamSpec.scala
```diff
@@ -65,6 +65,9 @@ class KernelInputStreamSpec
     // set of data
     doReturn(system.actorSelection(fakeInputOutputHandlerActor.path.toString))
       .when(mockActorLoader).load(MessageType.Incoming.InputReply)
+    // Allow time for the actors to start. This avoids read() hanging forever
+    // when running tests in gradle.
+    Thread.sleep(100)
```
rdblue: I do, too. The problem is that this test is flaky on my machine without it. We need to take a closer look at the flaky test, but I don't think this PR is the place to fix it. How about I open an issue for it and we solve the problem separately?

chipsenkbeil: Sounds good. Go with that.

kernel/src/test/scala/org/apache/toree/kernel/protocol/v5/relay/ExecuteRequestRelaySpec.scala
```diff
@@ -88,7 +89,7 @@ class ExecuteRequestRelaySpec extends TestKit(
         // Expected does not actually match real return of magic, which
         // is a tuple of ExecuteReply and ExecuteResult
         val expected = new ExecuteAborted()
-        interpreterActorProbe.expectMsgClass(
+        interpreterActorProbe.expectMsgClass(max = Duration(5, TimeUnit.SECONDS),
```
rdblue: The default is 3 seconds, but it was flaky for me. This was another work-around that probably shouldn't be here.

scala-interpreter/build.sbt
rdblue: I think we'll need to take a look after this PR at the other interpreters. PySpark, for example, should adhere to the common python convention to use _repr_html_ for objects that exist in the python process, and should use the Jupyter library for objects that are resident in the JVM (like DataFrames).

The background on Jupyter's jvm-repr library is that I'm trying to work with the Jupyter community to standardize a system for libraries and kernels to collaborate on rich representations for the JVM. Vegas, for example, should register a function to display its objects so it doesn't have to be built into the kernel or bolted on by users.

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaDisplayers.scala
rdblue: Calling this ensures the `ScalaDisplayers` object has been loaded, which performs the registration. Without calling this to make sure it loads, the registration further down never happens and we don't get representations.

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
rdblue: This makes sure that the registration in `ScalaDisplayers` happens when that class loads.

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaDisplayers.scala
chipsenkbeil: Can you add a doc to explain that reason? Otherwise, I might forget later and try to remove it.

scala-interpreter/build.sbt
chipsenkbeil: Given that all of the interpreter output gets funneled through the JVM before being sent back, would it make more sense for us to use this library at the JVM level for all of them? Either way, seems like another issue to open for discussion.

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaDisplayers.scala
rdblue: Will do.

scala-interpreter/build.sbt
rdblue: Yes, we probably should.

kernel-api/src/main/scala/org/apache/toree/interpreter/broker/BrokerTransformer.scala
```diff
@@ -43,7 +43,7 @@ class BrokerTransformer {
     import scala.concurrent.ExecutionContext.Implicits.global
 
     futureResult
-      .map(results => (Results.Success, Left(results)))
+      .map(results => (Results.Success, Left(Map("text/plain" -> results))))
```
rdblue: No, kernel-api doesn't depend on protocol so it isn't available. But I think that this code shouldn't be sending mime types in the first place. Sending text/plain here happens because we haven't updated the Python kernel to correctly inspect its objects.

Eventually, the python interpreter should send back a result that is a py4j reference to the final value and the python representation (from _repr_ and _repr_html_) or a JVM object and None to indicate that the Jupyter JVM displayer should be called. So for now, I'd rather not change module dependencies.

kernel/src/test/scala/org/apache/toree/kernel/protocol/v5/stream/KernelInputStreamSpec.scala
rdblue: Opened https://issues.apache.org/jira/browse/TOREE-416

kernel/src/test/scala/org/apache/toree/kernel/protocol/v5/relay/ExecuteRequestRelaySpec.scala
rdblue: Opened https://issues.apache.org/jira/browse/TOREE-417

kernel-api/src/main/scala/org/apache/toree/interpreter/broker/BrokerTransformer.scala
rdblue: Opened https://issues.apache.org/jira/browse/TOREE-418

rdblue:
@chipsenkbeil, I've updated this and opened the follow-up issues. If you think it is ready, I'll commit it.

I've not committed anything before. What is normal? I'd normally squash the commits first.

chipsenkbeil:
@rdblue, sure, go ahead. Normally, we would ask the owner of the PR to rebase first to have a clean merge, which allows the ASF bot to mark the PR as merged. If that's not possible, we would rebase and amend the last commit to close the PR. Squashing was optional, but I'm perfectly fine doing that to clean up the commits.

rdblue:
Merged! It was already based on the current master and I squashed the commits. Thanks for your help here.

I forgot to add the comment to close this, so I'll just do it manually.

PR #106, opened 2017-02-16
TOREE-388: Add gradle build.

rdblue:
Adds a gradle build script. The uses of `BuildInfo` have been removed because it is sbt-specific. It's possible to add a similar class with Gradle, if anyone is interested in using this.


chipsenkbeil:
@mariusvniekerk, @lbustelo, @Lull3rSkat3r, do you guys have any thoughts on this? I asked @rdblue to make an issue and PR so we can see if we'd like to include it.

Something I don't like is that our automated `BuildInfo` has been removed in favor of hard-coded values since Gradle cannot take advantage of the generated object from sbt. If we wanted to support a Gradle build, maybe we could have an external config file that Gradle and sbt can load and use when building? We could then either generate the `BuildInfo` class based on the config file or bundle the config file as a resource and change the `BuildInfo` class to read the config file at runtime.

Is anyone interested in this? Too much work to maintain the two? I know that Apache Spark had support for both Maven and sbt (I think they still do), but not sure if it's something we want to consider.

@rdblue, do you have any _pros_ to including this other than you and/or your team are using Gradle?

rdblue:
I agree that we would have to replace `BuildInfo`; I just didn't want to go to the trouble unless there was interest in a gradle-based build. My main motivation is to see if anyone else is interested so we don't do the same work twice, not necessarily to convince everyone to move to gradle.

The benefits I like are:
* Gradle appears to be faster. If I run from a cleaned directory with no gradle daemon running, the build with tests takes under 3 minutes. With sbt (make build test), the build takes just under 7 minutes.
* There is a single build script that is easy to follow. I find sbt builds hard to read, but gradle is reasonably straight-forward.
* Gradle is easier to use (in my opinion) and has good support for tasks like dependency inspection.

mariusvniekerk:
So we've recently refactored the sbt piece of toree.  

As long as this still produces the ivy XML pieces needed to play nice with coursier and dependency fetching I have no objections.

rdblue:
I don't know anything about the coursier requirements you mentioned, so I doubt it works. Are you talking about using Coursier for compile time dependency management, or for runtime downloads? We don't intend to have users download dependency trees a runtime, so I didn't really look into it.

chipsenkbeil:
@rdblue, we generate Ivy XML describing the different modules of Toree (kernel, kernel-api, communication, etc). The reason we do this is to use it with our dynamic dependency downloader used in `%%AddDeps`. We do this to avoid adding the same jar more than once for the compile classpath and runtime classloader used for interpreting Scala code.

The relevant code to generate the Ivy XML files is https://github.com/apache/incubator-toree/blob/master/project/CommonPlugin.scala#L47
The relevant code to load the Ivy XML files and use them for exclusions is https://github.com/apache/incubator-toree/blob/master/kernel-api/src/main/scala/org/apache/toree/dependencies/CoursierDependencyDownloader.scala#L286

This is purely for adding dependencies at runtime.

chipsenkbeil:
@rdblue FYI, I'm all for easier and faster. If you want to work on tackling the Ivy generation and making sure it plays well with our dependency downloaders, I'll try to pitch a little.

I'm at the airport right now and flying off to http://scalasphere.org for the week, but I'll be around some next week if needed.

rdblue:
I haven't been able to get the Ivy XML done yet -- sorry for the delay. I was able to generate it in the latest WIP commit, but I wasn't able to generate it at the right stage in the build where we can include it in the Jars. Maybe that's okay and we just need to include the XML in the final assembly Jar?

I also found that gradle doesn't have default support for scalatest, so I'll have to add that as well.

rdblue:
Okay, I finally was able to get the Ivy descriptors built and included in the jars. Here's a sample descriptor generated by gradle:

```
<?xml version="1.0" encoding="UTF-8"?>
<ivy-module version="2.0">
  <info organisation="org.apache.toree.kernel" module="kernel" revision="0.2.0" status="integration" publication="20170416153017"/>
  <configurations>
    <conf name="compile" visibility="public"/>
    <conf name="default" visibility="public" extends="runtime,compile"/>
    <conf name="runtime" visibility="public"/>
  </configurations>
  <publications>
    <artifact name="kernel" type="jar" ext="jar"/>
    <artifact name="kernel" type="jar" ext="jar" conf="runtime"/>
  </publications>
  <dependencies>
    <dependency org="org.apache.toree.kernel" name="macros" rev="0.2.0" conf="runtime-&gt;default"/>
    <dependency org="org.apache.toree.kernel" name="plugins" rev="0.2.0" conf="runtime-&gt;default"/>
    <dependency org="org.apache.toree.kernel" name="protocol" rev="0.2.0" conf="runtime-&gt;default"/>
    <dependency org="org.apache.toree.kernel" name="communication" rev="0.2.0" conf="runtime-&gt;default"/>
    <dependency org="org.apache.toree.kernel" name="kernel-api" rev="0.2.0" conf="runtime-&gt;default"/>
    <dependency org="org.apache.toree.kernel" name="pyspark-interpreter" rev="0.2.0" conf="runtime-&gt;default"/>
    <dependency org="org.apache.toree.kernel" name="scala-interpreter" rev="0.2.0" conf="runtime-&gt;default"/>
    <dependency org="org.apache.toree.kernel" name="sparkr-interpreter" rev="0.2.0" conf="runtime-&gt;default"/>
    <dependency org="org.apache.toree.kernel" name="sql-interpreter" rev="0.2.0" conf="runtime-&gt;default"/>
  </dependencies>
</ivy-module>
```

What is a good test case to verify that it prevents Coursier from downloading the Toree Jars?

I also updated the build to match the current sbt dependencies, run tests, and work with the latest gradle, 3.5. My timed runs before evidently didn't include running tests, so I'm not sure if this is actually much faster than sbt. Still, I think the gradle build is useful.

PR #107, opened 2017-02-27
[TOREE-386] Overridable application name

jodersky:
The application name of Toree's Spark driver program is currenlty
hardcoded to "Apache Toree" in the Spark context initialization
functions. By moving the name to a parameter into the launcher script,
it is possible for users to override the name by specifying an
additional "--name" parameter in the SPARK_OPTS variable (this is
because later parameters override earlier ones).

### DIFF
--- a/etc/bin/run.sh
+++ b/etc/bin/run.sh
@@ -42,4 +42,11 @@ then
    TOREE_OPTS=${__TOREE_OPTS__}
 fi
 
-eval exec "${SPARK_HOME}/bin/spark-submit" "${SPARK_OPTS}" --class org.apache.toree.Main "${TOREE_ASSEMBLY}" "${TOREE_OPTS}" "$@"
+eval exec \
+     "${SPARK_HOME}/bin/spark-submit" \
+     --name "Apache Toree"
+     "${SPARK_OPTS}" \
+     --class org.apache.toree.Main \
+     "${TOREE_ASSEMBLY}" \
+     "${TOREE_OPTS}" \
+     "$@"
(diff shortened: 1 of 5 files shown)


HySoaKa:
Will this correct the same issue on toree 0.1.0.dev8 for Spark 1.X ?

jodersky:
It won't work out-the-box for older versions; we'd need to back port it to 0.1.x branch.

@chipsenkbeil do you think this is something that could make it into 0.1.x?

mariusvniekerk:
So for 0.1 users can start with - - nosparkcontext and then construct their own sparkconf.  This allows setting the application name

HySoaKa:
@mariusvniekerk I was thinking about this solution. Thank you anyway!

jodersky:
The build failure seems to be unrelated (`make system-test` also fails on master) but unit tests pass.

lresende:
LGTM, thanks @jodersky

etc/bin/run.sh
```diff
@@ -42,4 +42,11 @@ then
    TOREE_OPTS=${__TOREE_OPTS__}
 fi
 
-eval exec "${SPARK_HOME}/bin/spark-submit" "${SPARK_OPTS}" --class org.apache.toree.Main "${TOREE_ASSEMBLY}" "${TOREE_OPTS}" "$@"
+eval exec \
+     "${SPARK_HOME}/bin/spark-submit" \
+     --name "Apache Toree"
```
ckadner: @jodersky -- missing a backslash to not break up the command

jodersky: fixed, thanks Christian

PR #108, opened 2017-03-04
TOREE-389: Ensure bindings are run before user code.

rdblue:


mariusvniekerk:
What does this change?

rdblue:
This binds `spark` before this runs:

```
    logger.info("Marking relay as ready for receiving messages")
    kernelMessageRelayActor ! true
```

That way if a cell was submitted during kernel startup, it can't run and fail. This prevents the interpreter hanging, which is a larger issue described on TOREE-389.

mariusvniekerk:
Ah, yeah that's a pain.

mariusvniekerk:
So the tests for this hung?  Dan you run these locally on your own system?  Alternatively closing and reopening will kick travis into restarting them.

PR #109, opened 2017-03-04
TOREE-390: Lazily initialize spark sessions

rdblue:
This updates the kernel so that it lazily initializes Spark sessions. This allows the user to start a configure and start a session, or to wait until `spark` is first referenced to start a session. This shortens the time to the first interaction with Toree.

The second commit adds a notification that Toree is waiting for a Spark session to start when the `getOrCreate` method takes more than 100ms.


kernel/src/main/scala/org/apache/toree/kernel/api/Kernel.scala
```diff
@@ -339,33 +340,8 @@ class Kernel (
...
-    val _sparkSession = SparkSession.builder.config(sconf).getOrCreate()
-
-    val sparkMaster = sconf.getOption("spark.master").getOrElse("not_set")
-    logger.info( s"Connecting to spark.master $sparkMaster")
-
-    // TODO: Convert to events
-    pluginManager.dependencyManager.add(_sparkSession.sparkContext.getConf)
-    pluginManager.dependencyManager.add(_sparkSession)
-    pluginManager.dependencyManager.add(_sparkSession.sparkContext)
-    pluginManager.dependencyManager.add(javaSparkContext(_sparkSession))
```
rdblue: Rather than converting these to events, plugins should reference the kernel and use `sparkSession` and similar methods to get access to Spark. That way only kernel is a dependency and the references passed to the dependency manager here don't get stale.

mariusvniekerk: Alternatively we could fire an event that users can subscribe to.

rdblue: Agreed. Either way, I don't think that this is necessary so I don't replace it in this patch.

rdblue:
@mariusvniekerk, what needs to happen to get this commit in?

kernel/src/main/scala/org/apache/toree/kernel/api/Kernel.scala
mariusvniekerk: Main other part is if we refactored toree to be less dependent on spark

mariusvniekerk:
So this seems to have affected the kernel tests? You can test locally with `make system-test`

rdblue:
I tried running the tests locally, but I get a strange error about ivy files:

```
[info] Compiling 2 Scala sources to /home/blue/workspace/toree/macros/target/scala-2.11/classes...
java.lang.RuntimeException: Duplicate mappings:
        /home/blue/workspace/toree/macros/target/scala-2.11/classes/toree-macros-ivy.xml
from
        /home/blue/workspace/toree/macros/target/scala-2.11/resource_managed/main/toree-macros-ivy.xml
                /home/blue/workspace/toree/macros/src/main/resources/toree-macros-ivy.xml
```

mariusvniekerk:
@jodersky Do you have an idea if this could be because of the changes that we made to the sbt things recently?

kernel/src/main/scala/org/apache/toree/kernel/api/Kernel.scala
```diff
@@ -396,7 +372,29 @@ class Kernel (
     interpreterManager.interpreters.get(name)
   }
 
-  override def sparkSession: SparkSession = SparkSession.builder.getOrCreate
+  private lazy val defaultSparkConf: SparkConf = createSparkConf(
+    new SparkConf().setAppName(SparkKernelInfo.banner))
```
kevin-bates: I don't think we want to set the appName here since it conflicts with this recent PR: https://github.com/apache/incubator-toree/pull/107.

kevin-bates: Btw, I think your build issue can be resolved via `git clean -xdf`.  This came up in the chat earlier: https://gitter.im/apache/toree?at=58c6cd0509e7ba8510bd50b0  I believe that's a one time thing following Jakob's sbt changes.

kevin-bates: I patched my fork with this PR and it looks good - although I did confirm it reverts the change in PR 107.  That said, I also confirmed that removing `.setAppName(SparkKernelInfo.banner)` restores the expected behavior and allows for --name overrides.

lresende: @rdblue Could you please take a quick look at the comments above, particularly for the setAppName then we will be getting ready to merge.

lresende:
LGTM with minor removal of .setAppName(SparkKernelInfo.banner)

rdblue:
I'll fix the name problem, but there's an additional problem with this before we can merge it. Some of the plugins like AddDeps and AddJar depend on the Spark context via IncludeSparkContext. I think this should remove that class if we are no longer going to send the events. Otherwise, the context is always null and AddDeps breaks.

lresende:
Thanks for the quick reply @rdblue , I will wait for the next updates from you then.

jodersky:
@kevin-bates @rdblue, could you see if the double-inclusion error persists after a `git clean`? Changes to the build occasionally cause `sbt clean` to miss things

kevin-bates:
Thanks @jodersky.  I'm confident `git clean` resolves that issue, although I'm not in a position to try it since its a one-time need following the pull of the sbt changes.  I mentioned this in an earlier comment in this PR and given it helped move @lbustelo forward for PR #111, I suspect @rdblue is good to go - but I'll have to defer to him.

rdblue:
I've updated this PR and removed the IncludeSpark* classes that are no longer notified since Spark is lazily initialized. These were only used for AddDeps and AddJars, which now pass the new Jar URIs to the kernel rather than working directly with Spark. Eventually, I think we should move to a SparkKernel implementation that layers on the Spark methods if it is enabled.

I also removed the appName call, so this no longer conflicts with #107.

mariusvniekerk:
@rdblue So the AddJarSpec test failures are probably the spark context not being around yet.

rdblue:
Yeah, I need to fix the tests for this one.

rdblue:
@mariusvniekerk, I fixed the scala tests, but now I have failures in Python. Any idea what those are? There's not much information to go on.

mariusvniekerk:
So I suspect what's happening here is that the integration test is picking up the notification when starting up spark. 

You probably want to adjust the test handler so that you filter out all the non-`execute_result` messages.

rdblue:
@mariusvniekerk, if I run the system-test locally from the commit where this branched, 6997e4d4e622e0af39e60f69417a4fe512b1b10c, I get the same test failures. I don't think they are related to this patch.

Also, we don't control those tests because they are based on [jupyter/jupyter_kernel_test](https://github.com/jupyter/jupyter_kernel_test/blob/309feed1670140bfd596ebde8bfc12e945a7199c/jupyter_kernel_test/__init__.py). I'll try rebasing this on master to see if that helps.

mariusvniekerk:
So we do pretty heavily adjust the kernel test pieces. I assume we are just getting messages in a different order.

lresende:
@rdblue On your side, is there any pending things for this PR  ? Are we just waiting on the build issue ?

rdblue:
Python tests are failing, but I haven't had time to really find out what's going on. I'm not sure that the test failures are caused by this patch... but I don't think we should commit it until those problems are solved.

lresende:
@rdblue I took a quick look at the PR, and added some more debug statements, and it seems that the failure is related to some timeouts waiting for the spark context to start:

This is what I am seeing on the messages received from the kernel
```

{'buffers': [],

  'content': {'data': {'text/plain': 'Waiting for a Spark session to start...'},

              'metadata': {},

              'source': 'user'},

  'header': {'date': '2017-05-30T00:54:07.076455',

             'msg_id': 'e74f827d-5694-463e-b71a-8d9b53b80a96',

             'msg_type': 'display_data',

             'session': '704f5820-eb0d-48e7-b088-f2ee0c2889f8',

             'username': 'root',

             'version': '5.0'},

  'metadata': {'timestamp': '1496105642479'},

  'msg_id': 'e74f827d-5694-463e-b71a-8d9b53b80a96',

  'msg_type': 'display_data',

  'parent_header': {'msg_id': 'e4b5bb9d-683e-4a3d-ad17-d0e75f202c5a',

                    'msg_type': 'execute_request',

                    'session': '44d3ab80-6bf8-45ff-a233-d03e507a6db4',

                    'username': 'username',

                    'version': '5.0'}},
```

And it seems it's failing in Kernel.scala

```
override def sparkSession: SparkSession = {
    // the first call to getOrCreate may create a session and take a long time,
    // so this starts a future to get the session. if it take longer than 100 ms,
    // then print a message to the user that Spark is starting.
    val sessionFuture = Future {
      SparkSession.builder.config(defaultSparkConf).getOrCreate
    }

    try {
      Await.result(sessionFuture, Duration(100, TimeUnit.MILLISECONDS))
    } catch {
      case timeout: TimeoutException =>
        // getting the session is taking a long time, so assume that Spark is
        // starting and print a message
        display.content(MIMEType.PlainText, "Waiting for a Spark session to start...")
        Await.result(sessionFuture, Duration.Inf)
    }
  }
```

rdblue:
@lresende, thanks! I recently found something similar (if not the same problem) in our deployment and added a work-around, so it makes sense... I'll make some time this week to get this updated.

rdblue:
@lresende, you were right. The problem was the notification when getting a Spark context. I updated the code to only send that notification when not running Spark in local mode, when starting a Spark context is quicker because you don't wait on a resource allocator. That fixes the tests.

Thanks for the help, @mariusvniekerk and @lresende! Think this is ready now?

rgbkrk:
Hurrah!

PR #112, opened 2017-03-11
Distinguish between shutdown request and System.exit from cell

kevin-bates:
Prior to this change there were three ways the toree process could be exited via deliberate actions. 

1. The user could enter dual cntrl-c (or send SIGINT) signals within 3 seconds of each other 
2. The user could issue a formal shutdown request from Jupyter 
3. The user could execute a cell of a notebook containing System.exit(n) or its equivalent 

The first two use cases are considered valid and are permitted.  Use case three, however, should be prevented by design.

The issue being addressed is a conflict between the second and third cases since both of those cases occur within the same thread group (referred to as the restricted group).  Prior to this change, shutdown requests made via Jupyter (vs. System.exit from within a cell) were not distinguished from cell requests.  Since the former is processed by the ShutdownHandler while the latter is processed by the ExecuteRequestHandler the change is to set a thread-local residing in the KernelSecurityManager from the ShutdownHandler immediately prior to its invocation of System.exit(0).  The KernelSecurityManager.checkExit() method then consults the thread local to ensure this call came by way of the ShutdownHandler and allows the exit to take place.  If the thread-local has not been set or has a value of false and its within the restricted group, a security exception is thrown to deny the exit attempt.

This topic was briefly discussed on gitter: https://gitter.im/apache/toree?at=58b9a97200c00c3d4fb6799f

### DIFF
--- a/kernel-api/src/main/scala/org/apache/toree/security/KernelSecurityManager.scala
+++ b/kernel-api/src/main/scala/org/apache/toree/security/KernelSecurityManager.scala
@@ -60,7 +60,7 @@ object KernelSecurityManager {
    * cases that don't exist in the normal name conventions.
    *
    * @param name The name of the permission
-   *:
+   *
    * @return True if the permission is to be checked, false otherwise
    */
   private def shouldCheckPermissionSpecialCases(name: String): Boolean =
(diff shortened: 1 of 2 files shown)


lresende:
LGTM, merging if no more comments

PR #116, opened 2017-03-21
[TOREE-395] disable cell result output from console

kevin-bates:
This issue - discussed in [TOREE-395](https://issues.apache.org/jira/browse/TOREE-395) (and somewhat in [TOREE-380](https://issues.apache.org/jira/browse/TOREE-380)) - essentially takes the appropriate portion of PR #104 and delivers it here.  This change makes the scala interpreter behave like other interpreters (and kernels) by not issuing the cell results to the console (where it is typically captured by the Jupyter logging) and could contain sensitive information.

### DIFF
--- a/scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
+++ b/scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
@@ -48,7 +48,7 @@ class ScalaInterpreter(private val config:Config = ConfigFactory.load) extends I
 
    protected val lastResultOut = new ByteArrayOutputStream()
 
-   protected val multiOutputStream = MultiOutputStream(List(Console.out, lastResultOut))
+   protected val multiOutputStream = lastResultOut
    private[scala] var taskManager: TaskManager = _
 
   /** Since the ScalaInterpreter can be started without a kernel, we need to ensure that we can compile things.
--- a/scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
+++ b/scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
@@ -29,7 +29,7 @@ import org.apache.spark.repl.Main
 
 import org.apache.toree.interpreter._
 import org.apache.toree.kernel.api.{KernelLike, KernelOptions}
-import org.apache.toree.utils.{MultiOutputStream, TaskManager}
+import org.apache.toree.utils.TaskManager
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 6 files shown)


rdblue:
@kevin-bates, looks like there is a compile failure. Can you fix it and push an update?

kevin-bates:
Thanks @rdblue.

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
```diff
@@ -48,7 +48,7 @@ class ScalaInterpreter(private val config:Config = ConfigFactory.load) extends I
 
    protected val lastResultOut = new ByteArrayOutputStream()
 
-   protected val multiOutputStream = MultiOutputStream(List(Console.out, lastResultOut))
+   protected val multiOutputStream = MultiOutputStream(List(lastResultOut))
```
rdblue: I think the multiOutputStream can be completely removed rather than using it to wrap just one output stream. Instead, lastResultOut should be passed to IMain instead of multiOutputStream.

kevin-bates: multiOutputStream is overridden by a test class that requires the list, that was the cause of the compilation failure. Your PR should have the same failure.

chipsenkbeil:
Should this be configurable? Maybe through a command line option? Or do you guys think that the `Console.out` channel should be removed entirely?

kevin-bates:
I guess I can see some benefit for debugging - especially as we get into remote kernel scenarios (which is happening).  However, I've also been around enough to know that even the slightest existence of sensitive data persisted in places it shouldn't go can be considered a serious security concern.  Since we have no control over what an end user wishes to output in their notebook, then I'm more of the opinion that we should not permit its capture at all. 
I don't believe other kernels/interpreters (including those in Toree) perform this kind of capture either.  So for that there's precedent as well.

scala-interpreter/src/main/scala/org/apache/toree/kernel/interpreter/scala/ScalaInterpreter.scala
rdblue: I'd rather change the test to fix this than keep a variable that is useless. Is that possible?

kevin-bates: Yes - testing a change now. It just trickled to a few test files and is looking good.

kevin-bates:
close/reopen to trigger build - looks like the travis server went down mid-stream.

rdblue:
+1 (non-binding)

lresende:
LGTM, merging if there are no more comments...

chipsenkbeil:
Go ahead and merge.

PR #118, opened 2017-03-23
Allow also URLs with additional parameters

kamir:
Not all URLs end with JAR or ZIP, e.g., if we load content from an HTTPFS Service.

Examples for testing: (a,c,d work as expeced) (b => more logic needed to create a JAR name from this URL)

import java.io.{File, PrintStream}
import java.net.URL
import java.nio.file.{Files, Paths}

def getFileFromLocation(location: String): String = {
    val url = new URL(location)
    val file = url.getFile.split("/")
    if (file.length > 0) {
        if ( file.last.contains('?') ) {
          file.last.split('?')(0)
        }
        else { 
          file.last
        }  
    } else {
        ""
    }
  }

val fnA = "http://cdsw-mk4-1.gce.cloudera.com:8888/filebrowser/download=/user/systest/TOOLBOX/SparkShellUtilities.jar?user.name=123"
val a = getFileFromLocation(fnA)
a
  
val fnB = "https://oss.sonatype.org/service/local/artifact/maven/content?r=snapshots&g=org.gephi&a=gephi-toolkit&v=0.9.2-SNAPSHOT&c=all"
val b = getFileFromLocation(fnB)
b

val fnC = "http://cdsw-mk4-1.gce.cloudera.com:8888/filebrowser/download=/user/systest/TOOLBOX/SparkShellUtilities.jar"
val c = getFileFromLocation(fnC)
c
  
val fnD = "http://cdsw-mk4-1.gce.cloudera.com:8888/filebrowser/download=/user/systest/TOOLBOX/SparkShellUtilities.jar?user.name=123&doas=123456"
val d = getFileFromLocation(fnD)
d

### DIFF
--- a/kernel/src/main/scala/org/apache/toree/magic/builtin/AddJar.scala
+++ b/kernel/src/main/scala/org/apache/toree/magic/builtin/AddJar.scala
@@ -67,11 +67,16 @@ class AddJar
    * @param location The remote location (URL) 
    * @return The name of the remote URL, or an empty string if one does not exist
    */
-  def getFileFromLocation(location: String): String = {
+   def getFileFromLocation(location: String): String = {
     val url = new URL(location)
     val file = url.getFile.split("/")
     if (file.length > 0) {
-        file.last
+        if ( file.last.contains('?') ) {
+          file.last.split('?')(0)
+        }
+        else { 
+          file.last
+        }  
     } else {
         ""
     }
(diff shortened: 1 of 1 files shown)


lresende:
@kamir  Could you please update AddJarSpec.scala with some tests that validate your additions.

jodersky:
Welcome @kamir! Thanks for finding this bug and working on a fix. Check out this documentation on URL https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#getFile(), I think a simpler fix is to just change the "getFile" part above to "getPath"

lresende:
@kamir Could you please rebase and add a test case to the enhancements you are making.

kamir:
My plan is to work on this during the next week.
Best wishes,
Mirko

PR #120, opened 2017-06-10
TOREE-412: Update AddDeps magic

rdblue:
* Removes non-Jar files (POM files) from downloaded artifacts before adding dependencies to the kernel
* Updates to Coursier 1.0.0-M15-7
* Uses "default" configuration by default, instead of "compile" for Ivy dependencies
* Adds `--classifier` for maven artifacts
* Adds `--ivy-configuration` for Ivy artifacts

### DIFF
--- a/kernel-api/src/main/scala/org/apache/toree/dependencies/CoursierDependencyDownloader.scala
+++ b/kernel-api/src/main/scala/org/apache/toree/dependencies/CoursierDependencyDownloader.scala
@@ -20,16 +20,14 @@ import java.io.{File, FileInputStream, PrintStream}
 import java.net.{URI, URL}
 import java.util.Properties
 import java.util.concurrent.ConcurrentHashMap
-
 import coursier.core.Authentication
 import coursier.Cache.Logger
 import coursier.Dependency
 import coursier.core.Repository
-import coursier.core.Resolution.ModuleVersion
 import coursier.ivy.{IvyRepository, IvyXml}
 import coursier.maven.MavenRepository
 import org.springframework.core.io.support.PathMatchingResourcePatternResolver
-
+import scala.util.Try
 import scalaz.\/
 import scalaz.concurrent.Task
 
@@ -117,7 +115,7 @@ class CoursierDependencyDownloader extends DependencyDownloader {
     val resolution = start.process.run(fetch).unsafePerformSync
 
     // Report any resolution errors
-    val errors: Seq[(ModuleVersion, Seq[String])] = resolution.metadataErrors
+    val errors: Seq[((Module, String), Seq[String])] = resolution.metadataErrors
     errors.foreach { case (d, e) =>
       printStream.println(s"-> Failed to resolve ${d._1.toString()}:${d._2}")
       e.foreach(s => printStream.println(s"    -> $s"))
@@ -313,10 +311,11 @@ class CoursierDependencyDownloader extends DependencyDownloader {
    * @param repositories The repositories to convert
    * @return The resulting URIs
    */
-  private def repositoriesToURIs(repositories: Seq[Repository]) = repositories.map {
-    case ivy: IvyRepository => ivy.pattern.string
-    case maven: MavenRepository => maven.root
-  }.map(new URI(_))
+  private def repositoriesToURIs(repositories: Seq[Repository]) =
+    repositories.map {
+      case ivy: IvyRepository => ivy.pattern.string
+      case maven: MavenRepository => maven.root
+    }.map(s => Try(new URI(s))).filter(_.isSuccess).map(_.get)
 
   /** Creates new Ivy2 local repository using base home URI. */
   private def ivy2Local(ivy2HomeUri: URI) = IvyRepository.parse(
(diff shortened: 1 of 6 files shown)


kalvinnchau:
I pulled in the interpreter changes which solved an issue where I would get:
```
 found: org.apache.spark.org.apache.spark.org.apache.spark.org.apache.spark.org.apache.spark.SparkContext'
required: org.apache.spark.org.apache.spark.org.apache.spark.org.apache.spark.org.apache.spark.SparkContext'
```
After running the AddDeps magic.  Thanks for this! Good work

rdblue:
@kalvinnchau, that was fixed by "Fix kernel references after %AddDeps." The problem was that the interpreter was redefining the classes, but using the old reference. Rebinding everything after changing dependencies is what fixes it. Glad it is working for you.

kalvinnchau:
@rdblue Not sure if you've seen something really weird like this but I get a bunch of
```
error: missing or invalid dependency detected while loading class file 'SequenceFileRDDFunctions.class'.
Could not access type AnyRef in package scala,
because it (or its dependencies) are missing. Check your build definition for
missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
```

When running in a notebook
```
[1]
%AddDeps com.github.cb372 scalacache-caffeine_2.11 0.9.3 --transitive
%AddDeps org.scalaj scalaj-http_2.11 2.3.0 --transitive

[2]
import scalaj.http.{Http, Token}
val t = Token("key","secret")

[3]
val rdd = sc.parallelize(1 to 100)

[4]
val http = Http("http://google.com")
http = HttpRequest(http://google.com,GET,DefaultConnectFunc,List(),List((User-Agent,scalaj-http/1.0)),List....
error: missing or invalid dependency detected while loading class file 'SequenceFileRDDFunctions.class'.
Could not access type AnyRef in package scala,
because it (or its dependencies) are missing. Check your build definition for
missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
```


But when I do this:
```
[1]
%AddDeps com.github.cb372 scalacache-caffeine_2.11 0.9.3 --transitive
%AddDeps org.scalaj scalaj-http_2.11 2.3.0 --transitive

[2]
import scalaj.http.{Http, Token}
val t = Token("key","secret")

[3]
val rdd = sc.parallelize(1 to 100)
val new_token = Token("key","secret")

[4]
val http = Http("http://google.com")
http = HttpRequest(http://google.com,GET,DefaultConnectFunc,List(),List((User-Agent,scalaj-http/1.0)),List....
```

Or any variation of using one of the imports inside the same cell as doing some spark action, the errors go away.

Fairly easy to work around for now, just wondering if you'd have any idea why this happens

rdblue:
@kalvinnchau, Does applying this PR have any effect on the problem you just posted? That's really strange.

kalvinnchau:
@rdblue applying this PR allowed me to get that far :) haha

rdblue:
@chipsenkbeil may have an idea what is happening.

chipsenkbeil:
Couldn't tell you without diving deeper. I believe we are executing line-by-line when submitting code cells; so, I'm not sure why including it fixes that. @rdblue, none of your PRs changed our code execution behavior, did they? If we were bundling multiple lines together, it could affect the result based on how Scala's REPL generates the code to be evaluated.

rdblue:
I did change the behavior to evaluate blocks instead of individual lines: https://github.com/apache/incubator-toree/commit/5cd1727089eff26f553f990ab9dfe49d2be14fe2#diff-1008ba2d7bf6341cdaf51008230245c5R165.

That change handles the return value of a block as the object to display. It also avoids calling the compiler multiple times and, I think, handles cases slightly better where a statement appears to be complete but isn't (e.g., method chaining). What is the rationale for interpreting line-by-line?

chipsenkbeil:
When invoking line magics, we wanted to support you having `%AddJar` in the same cell as code that depended on the jar. With the line-by-line evaluation, the code would work as it would invoke the `%AddJar` separately from the latter code. When evaluating all code at once, it would fail because it could not compile the cell as the needed jar for the compile and runtime requirements had yet to be downloaded.

It's a minor case and I agree that cell-based evaluation feels a little more stable.

rdblue:
The tests caught the line magic issue. I think since those are already parsed out of the cell, we can handle it as a special case. I already want to avoid the sending all magic execution through the Scala interpreter, so we can fix %AddDeps when we fix that.

Calling magics directly will allow for better interaction with interpreters that are wrapped by magic. For example, the SQL interpreter can't use magics at all right now, and %%sql can't use the SQL interpreter's isComplete logic.

chipsenkbeil:
Agreed. I had wanted to move magics out of the interpreter route, but it wasn't a high enough priority at the time.

lresende:
@chipsenkbeil any other comments here?

@rdblue  Could you please rebase to latest master.

rdblue:
Rebased

chipsenkbeil:
No other comments.

thak123:
--classifier doesnt work ryt ?

PR #123, opened 2017-06-10
TOREE-415: Cancel Spark jobs when interrupted.

rdblue:


chipsenkbeil:
This is something we had planned to do a long time ago using the `SparkContext.setJobGroup` where each block of code being executed would set a new job group. Should this be a configurable option when starting the kernel? I'm also not sure what the ramifications of setting job groups over and over would be on Spark at this point.

Have either a) interrupt kill all Spark work, b) interrupt kill the last Spark job, or c) interrupt only stop current local code execution.

rdblue:
Spark's `setJobGroup` just sets an inherited variable in thread-local context. It should be safe to call multiple times to track jobs started by threads reliably. I don't think we need to do that just yet, since we currently don't have concurrent cell execution. The only way users would accidentally kill Spark jobs is if they were opting to add threads and sent an interrupt. I think it would be reasonable to put off the 100% solution in favor of the 80% one right now, since the current behavior where we don't cancel anything is worse. Still, if you think the 100% solution is important we can probably get that done without too much trouble.

lresende:
LGTM, Also agree with the 80% solution for now. Merging if there are no more comments.

mariusvniekerk:
Agree with @lbustelo here.  We should get this in. 

_Sent from my Google Pixel using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #125, opened 2017-06-10
TOREE-408: Add support for hdfs and s3 to AddJar.

rdblue:


liukun1016:
How to know the reason for the failure of the test? The details said 1 Test Failed for `org.apache.toree.magic.builtin.AddJarSpec` but no further explanations.

jodersky:
It's a bit further up, [line 3324](https://travis-ci.org/apache/incubator-toree/builds/241601768#L3324).

rdblue:
Rebased on master and fixed the test failure.

The problem was with the test itself. The addJar code was returning the file name based on a query param instead of the URL's path. Since query params can be in any order and can't be relied upon for correct file names, the right thing to do is to get the filename from the path component. I updated the test to do that.

rdblue:
@kamir, I think this fixes #118 because URI correctly parses the file name without query params. Can you have a look?

lresende:
LGTM, merging if there are no more comments.

PR #129, opened 2017-08-09
Fix for TOREE-430

Myllyenko:
To fix TOREE-430, REPL-generated classes must be made visible to a classloader.

The only working way I found is to set the `spark.repl.class.outputDir` system property (and subsequently the corresponding Spark configuration property) to a path to an output directory of a REPL compiler. To achieve that, I had to move `SparkIMain` initialisation before creation of `SparkContext`.

### DIFF
--- a/scala-interpreter/src/main/scala-2.10/org/apache/toree/kernel/interpreter/scala/ScalaInterpreterSpecific.scala
+++ b/scala-interpreter/src/main/scala-2.10/org/apache/toree/kernel/interpreter/scala/ScalaInterpreterSpecific.scala
@@ -61,6 +61,7 @@ trait ScalaInterpreterSpecific { this: ScalaInterpreter =>
   ): SparkIMain = {
     val s = new SparkIMain(settings, out)
     s.initializeSynchronous()
+    System.setProperty("spark.repl.class.outputDir", s.getClassOutputDirectory.getAbsolutePath)
     s
   }
 
@@ -288,6 +289,10 @@ trait ScalaInterpreterSpecific { this: ScalaInterpreter =>
     else variable
   }
 
+  protected def initializeIMain() = {
+    sparkIMain = newSparkIMain(settings, new JPrintWriter(lastResultOut, true))
+  }
+
   /**
    * Starts the interpreter, initializing any internal state.
    * You must call init before running this function.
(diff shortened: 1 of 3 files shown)


scala-interpreter/src/main/scala-2.10/org/apache/toree/kernel/interpreter/scala/ScalaInterpreterSpecific.scala
```diff
@@ -61,6 +61,7 @@ trait ScalaInterpreterSpecific { this: ScalaInterpreter =>
   ): SparkIMain = {
     val s = new SparkIMain(settings, out)
     s.initializeSynchronous()
+    System.setProperty("spark.repl.class.outputDir", s.getClassOutputDirectory.getAbsolutePath)
```
lresende: Could we pass this as conf while creating the Spark Session?

Myllyenko: I've tried to figure out how to do that without heavy code modifications but haven't succeeded.

It worth noting that currently in a Scala 2.11 environment Toree [sets this property in a similar manner](https://github.com/apache/incubator-toree/blob/master/kernel/src/main/scala/org/apache/toree/boot/KernelBootstrap.scala#L70). And in Scala 2.11 we can pass it directly to `SparkConf` with ease because `spark.repl.class.outputDir`'s required value becomes known very early ([when `org.apache.spark.repl.Main` is initialised](https://github.com/apache/spark/blob/branch-1.6/repl/scala-2.11/src/main/scala/org/apache/spark/repl/Main.scala#L33)).

But in a Scala 2.10 environment, things are different. `spark.repl.class.outputDir`' [is initialised only with `org.apache.spark.repl.SparkIMain`](https://github.com/apache/spark/blob/branch-2.2/repl/scala-2.10/src/main/scala/org/apache/spark/repl/SparkIMain.scala#L100), that is after creation of `ScalaInterpreter` in Toree. But in order to pass `spark.repl.class.outputDir` directly to `SparkConf` we have to initialise`SparkIMain` earlier than Toree's `Kernel`.

This is applicable for both Spark 2 and Spark 1.6.

lresende:
@Myllyenko Could you please rebase and update your pr so we can make sure we have a clean build before merging the update.

Myllyenko:
I started to work on an improved version of this modification (with more accurate splitting between 2.10/2.11 specifics). Unfortunately I didn't have enough time to finish it yet, but I hope it will be ready soon.

Myllyenko:
I've rebased this PR.
I'll open new PRs for the modifications I mentioned in my previous comment.

Myllyenko:
This PR is no longer relevant as it is Scala 2.10-specific.

PR #134, opened 2017-08-18
Fix null pointer in file utils

liukun1016:
Found this bug on my local dev env. I added the FileUtils on April.


lresende:
@liukun1016 It seems like you are reusing an old branch? if so, could you please rebase your master, and recreate this PR (or a new one) with only the changes required to fix the NPE.

PR #135, opened 2017-08-22
Check if file is null before delete on FileUtils

liukun1016:
Found this bug on my local dev env. I added the FileUtils on April but didn't check the null pointer exception then.


lresende:
@liukun1016  Could you please add a test case where you actually pass a null and an invalid directory.

lresende:
@liukun1016  ping

liukun1016:
@lresende will take a look at it tomorrow

PR #138, opened 2017-09-04
Fix for TOREE-396

lammic:
Added timeout parameter to call to `connectBackend`. Tested with Spark 2.2.0.


lresende:
LGTM, merging if no more comments

PR #140, opened 2017-09-11
[TOREE-437] Establish alternate interrupt handler for background execution

kevin-bates:
Cell interrupts do not take place if Toree is running in the background (typically due to its parent being in the background).  This PR adds support for the caller of Toree to specify an alternate interrupt signal via the environment variable `TOREE_ALTERNATE_SIGINT` - the value of which should be the string indicating the alternate signal name's suffix portion (e.g., "USR2" for `SIGUSR2`).
The common scenario would be to specify this env value in the kernel.json's `env:` stanza when necessary ...
```json
  "display_name": "Apache Toree - Scala",
  "env": {
    "DEFAULT_INTERPRETER": "Scala",
    "PYTHON_EXEC": "python",
    "__TOREE_SPARK_OPTS__": "",
    "__TOREE_OPTS__": "",
    "TOREE_ALTERNATE_SIGINT": "USR2",
```

If the value of `TOREE_ALTERNATE_SIGINT` is not recognized as a valid signal name, a warning message will be logged, but the kernel's launch will proceed:
```
17/09/11 08:44:37 [WARN] o.a.t.Main$$anon$1 - Error occurred establishing alternate signal handler.  Value of TOREE_ALTERNATE_SIGINT is probably bad: FOO.  Error: Unknown signal: FOO
```

Fixes TOREE-437

### DIFF
--- a/kernel/src/main/scala/org/apache/toree/boot/layer/HookInitialization.scala
+++ b/kernel/src/main/scala/org/apache/toree/boot/layer/HookInitialization.scala
@@ -69,8 +69,7 @@ trait StandardHookInitialization extends HookInitialization {
     // TODO: Signals are not a good way to handle this since JVM only has the
     // proprietary sun API that is not necessarily available on all platforms
     Signal.handle(new Signal("INT"), new SignalHandler() {
-      private val MaxSignalTime: Long = 3000
-      // 3 seconds
+      private val MaxSignalTime: Long = 3000 // 3 seconds
       var lastSignalReceived: Long = 0
 
       def handle(sig: Signal) = {
@@ -109,8 +108,8 @@ trait StandardHookInitialization extends HookInitialization {
             }
         })
       } catch {
-        case e:Exception => logger.warn("Error occurred establishing alternate signal handler.  Value of " +
-          "TOREE_ALTERNATE_SIGINT is probably bad: " + altSigint + ".  Error: " + e.getMessage )
+        case e:Exception => logger.warn("Error occurred establishing alternate signal handler " +
+          "(TOREE_ALTERNATE_SIGINT = " + altSigint + ").  Error: " + e.getMessage )
       }
     }
   }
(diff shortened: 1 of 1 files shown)


kernel/src/main/scala/org/apache/toree/boot/layer/HookInitialization.scala
```diff
@@ -69,8 +69,9 @@ trait StandardHookInitialization extends HookInitialization {
     // TODO: Signals are not a good way to handle this since JVM only has the
     // proprietary sun API that is not necessarily available on all platforms
     Signal.handle(new Signal("INT"), new SignalHandler() {
-      private val MaxSignalTime: Long = 3000 // 3 seconds
-      var lastSignalReceived: Long    = 0
+      private val MaxSignalTime: Long = 3000
+      // 3 seconds
+      var lastSignalReceived: Long = 0
```
lresende: Can we remove the formatting changes above?

kernel/src/main/scala/org/apache/toree/boot/layer/HookInitialization.scala
```diff
@@ -89,6 +90,29 @@ trait StandardHookInitialization extends HookInitialization {
...
+              logger.info("Resetting code execution due to interrupt!")
+              interpreter.interrupt()
+
+              // TODO: Cancel group representing current code execution
+              //sparkContext.cancelJobGroup()
+            }
+        })
+      } catch {
+        case e:Exception => logger.warn("Error occurred establishing alternate signal handler.  Value of " +
+          "TOREE_ALTERNATE_SIGINT is probably bad: " + altSigint + ".  Error: " + e.getMessage )
```
lresende: should it say 'bad' or some more specific message, or even no message and assume more details coming from the exception message?

lresende:
Other than minor formatting/error message comments, it LGTM.
Merging after the comments are addressed and no other feedback is provided.

kevin-bates:
Thanks @lresende - I've updated per your suggestions.  Decided to keep `TOREE_ALTERNATE_SIGINT` in warning message to assist with troubleshooting, but agree with your comment that previous message was poorly worded/unhelpful.

PR #142, opened 2017-09-26
add Python3 support

aleozlx:
This supports both Python2 and Python3 now. Just remember to export PYTHON_EXEC=python3 for example. You are welcome..


lresende:
Obsolete

PR #144, opened 2017-11-06
TOREE-454: Add RAT script.

rdblue:


lresende:
@rdblue How is this different from ```make audit-licenses``` that will call into etc/tools/check-licenses script ? And the rat exclusion is available as etc/tools/.rat-excludes.

And yes, I understand there is too many undocumented magic going on with these make infrastructure.

rdblue:
I guess it isn't. I didn't know about the other check.

PR #126, opened 2017-06-22
TOREE-420 Exclude org.scala-lang.modules from AddDeps

kalvinnchau:
https://issues.apache.org/jira/browse/TOREE-420

Include `org.scala-lang.modules` as part of the always exclude list when using the `%AddDeps` magic.

This prevents the `scala.reflect.internal.FatalError` that crashes the kernel, when a new scala-lang module is included and reloaded.

PR #146, opened 2017-11-15
Add --exclude option to AddDeps and exclude org.scala-lang.modules by default

kervel:
This PR includes the (in my eyes valid) fix to TOREE-420 by Kalvin Chau, and extends on it by adding a way to work around other dependency conflicts by excluding specific dependencies. This is already possible in other scala-based interactive notebooks.

For example, look at https://zeppelin.apache.org/docs/0.6.2/manual/dependencymanagement.html

### DIFF
--- a/kernel-api/src/main/scala/org/apache/toree/dependencies/CoursierDependencyDownloader.scala
+++ b/kernel-api/src/main/scala/org/apache/toree/dependencies/CoursierDependencyDownloader.scala
@@ -84,7 +84,7 @@ class CoursierDependencyDownloader extends DependencyDownloader {
     // Grab exclusions using base dependencies (always exclude scala lang)
     val exclusions: Set[(String, String)] = (if (excludeBaseDependencies) {
       getBaseDependencies.map(_.module).map(m => (m.organization, m.name))
-    } else Nil).toSet ++ Set(("org.scala-lang", "*"))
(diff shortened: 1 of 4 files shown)


kervel:
When #126 is merged i'll rebase this PR.

kervel:
no CI ?

lresende:
ok to test

kernel-api/src/main/scala/org/apache/toree/dependencies/CoursierDependencyDownloader.scala
```diff
@@ -76,15 +76,16 @@ class CoursierDependencyDownloader extends DependencyDownloader {
     trace: Boolean,
     configuration: Option[String] = None,
     artifactType: Option[String] = None,
-    artifactClassifier: Option[String] = None
+    artifactClassifier: Option[String] = None,
+    excludes: Set[(String,String)] = Set.empty
```
lresende: As you are adding a new parameter here, you also need to update the addDepsSpec where it defines a mock of a dependency downloader and add a new parameter to the mock (any[Set[(String,String)]]). I am playing with it and if that's the only thing I might update your pr before merging, otherwise please make sure you look into this.

lresende:
After the changes to AddDepsSpec.scala, I was able to get a [successful](https://travis-ci.org/lresende/incubator-toree/builds/328630546) build.

lresende:
LGTM, merging if there are no more comments

kervel:
thanks !

PR #147, opened 2017-11-20
Support for --spark-context-initialization-timeout command-line option

sanjay-saxena:
This change addresses TOREE-443. Instead of hardcoding the
timeout to 100ms, --spark-context-initialization-timeout
command-line option has been introduced to make it configurable.
On slow/busy nodes, an appropriate timeout can be specified
to avoid the confusing messages in the notebook UI. Added
unit tests to CommandLineOptionsSpec.scala.

### DIFF
--- a/kernel/src/main/scala/org/apache/toree/boot/CommandLineOptions.scala
+++ b/kernel/src/main/scala/org/apache/toree/boot/CommandLineOptions.scala
@@ -19,9 +19,7 @@ package org.apache.toree.boot
 
 import java.io.{File, OutputStream}
 
-import org.apache.toree.utils.KeyValuePairUtils
 import com.typesafe.config.{Config, ConfigFactory}
-import joptsimple.util.KeyValuePair
 import joptsimple.{OptionParser, OptionSpec}
 
 import scala.collection.JavaConverters._
@@ -100,6 +98,13 @@ class CommandLineOptions(args: Seq[String]) {
     "interpreter-plugin"
   ).withRequiredArg().ofType(classOf[String])
 
+  private val _spark_context_intialization_timeout = parser.accepts(
+    "spark-context-intialization-timeout",
+    "The time (in milliseconds) allowed for creation of the spark context. " +
+      "Failure to create a context in this time could result in duplicate initialization messages. " +
+      "The default value is 100 milliseconds."
+  ).withRequiredArg().ofType(classOf[Long])
+
   private val options = parser.parse(args.map(_.trim): _*)
 
   /*
@@ -149,6 +154,7 @@ class CommandLineOptions(args: Seq[String]) {
       "magic_urls" -> getAll(_magic_url).map(_.asJava)
         .flatMap(list => if (list.isEmpty) None else Some(list)),
       "max_interpreter_threads" -> get(_max_interpreter_threads),
+      "spark_context_intialization_timeout" -> get(_spark_context_intialization_timeout),
       "jar_dir" -> get(_jar_dir),
       "default_interpreter" -> get(_default_interpreter),
       "nosparkcontext" -> (if (has(_nosparkcontext)) Some(true) else Some(false)),
--- a/kernel/src/main/scala/org/apache/toree/kernel/api/Kernel.scala
+++ b/kernel/src/main/scala/org/apache/toree/kernel/api/Kernel.scala
@@ -392,6 +392,17 @@ class Kernel (
     sparkContext
   }
 
+  // TODO: Exposed for testing purposes.
+  protected[toree] def getSparkContextInitializationTimeout: Long = {
+    val timeout:Long = config.getDuration("spark_context_intialization_timeout", TimeUnit.MILLISECONDS)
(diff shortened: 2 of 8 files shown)


kernel/src/main/scala/org/apache/toree/kernel/api/Kernel.scala
```diff
@@ -403,15 +403,17 @@ class Kernel (
...
         // message to the user that Spark is starting.
         import scala.concurrent.ExecutionContext.Implicits.global
         val sessionFuture = Future {
           SparkSession.builder.config(defaultSparkConf).getOrCreate
         }
 
         try {
-          Await.result(sessionFuture, Duration(100, TimeUnit.MILLISECONDS))
+          val timeout = config.getInt("spark_context_intialization_timeout")
+          logger.info("Spark Context Initialization Timeout is {} milliseconds", timeout)
```
rdblue: This should not be logged.

kernel/src/main/scala/org/apache/toree/boot/CommandLineOptions.scala
```diff
@@ -193,6 +199,19 @@ class CommandLineOptions(args: Seq[String]) {
     Some(userDefined.asJava)
   }
 
+  private def onlyPositiveOtherwiseDefault(spec: OptionSpec[Int]): Option[Int] = {
```
rdblue: I don't think it is a good idea to silently ignore user input. If the value passed is -1, for example, then the user gets 100ms. If what the user requests isn't possible, then this should fail and print the help message.

kernel/src/main/scala/org/apache/toree/boot/CommandLineOptions.scala
```diff
@@ -100,6 +98,13 @@ class CommandLineOptions(args: Seq[String]) {
     "interpreter-plugin"
   ).withRequiredArg().ofType(classOf[String])
 
+  private val _spark_context_intialization_timeout = parser.accepts(
+    "spark-context-intialization-timeout",
+    "The time (in milliseconds) allowed for creation of the spark context. " +
+      "Failure to create a context in this time could result in duplicate initialization messages. " +
+      "The default value is 100 milliseconds."
+  ).withRequiredArg().ofType(classOf[Int]).defaultsTo(100)
```
rdblue: This should default to a configuration value, not a hard-coded 100ms. We can set the 100ms default in Toree's default config. That way this behaves like other options.

kernel/src/main/scala/org/apache/toree/kernel/api/Kernel.scala
```diff
@@ -403,15 +403,17 @@ class Kernel (
...
+        // to get the session. if it take longer than specified timeout, then print a
         // message to the user that Spark is starting.
         import scala.concurrent.ExecutionContext.Implicits.global
         val sessionFuture = Future {
           SparkSession.builder.config(defaultSparkConf).getOrCreate
         }
 
         try {
-          Await.result(sessionFuture, Duration(100, TimeUnit.MILLISECONDS))
+          val timeout = config.getInt("spark_context_intialization_timeout")
```
rdblue: There are accessors for time values that should be used here.

kernel/src/main/scala/org/apache/toree/boot/CommandLineOptions.scala
sanjay-saxena: Sounds good.

kernel/src/main/scala/org/apache/toree/boot/CommandLineOptions.scala
sanjay-saxena: I see the pattern in `resources/compile/application.conf` and `resources/compile/reference.conf` for other options with default values. I will just follow the pattern for the new option.

kernel/src/main/scala/org/apache/toree/kernel/api/Kernel.scala
sanjay-saxena: Sounds good.

kernel/src/main/scala/org/apache/toree/kernel/api/Kernel.scala
sanjay-saxena: I will remove the logging statements.

kernel/src/main/scala/org/apache/toree/kernel/api/Kernel.scala
```diff
@@ -403,15 +414,16 @@ class Kernel (
       case Some(master) if !master.contains("local") =>
         // when connecting to a remote cluster, the first call to getOrCreate
         // may create a session and take a long time, so this starts a future
-        // to get the session. if it take longer than 100 ms, then print a
+        // to get the session. if it take longer than specified timeout, then print a
```
lresende: maybe you should still mention that (100 ms is default)

sanjay-saxena: Updated the comment.

PR #148, opened 2017-12-05
Add support for showing nulls and arrays by dataframe magic

k1s:
This change adds support for showing dataframes with arrays and nulls by %%dataframe magic. Now null values cause NullPointerException and arrays are showed as "WrappedArray(1, 2, 3)" but not "[1, 2, 3]".


k1s:
Test that failed fails on master too, but when I run it in sbt everything is fine, is that correct?

lresende:
LGTM

@k1s I have created [TOREE-462](https://issues.apache.org/jira/browse/TOREE-462), could you please let me know your jira user or create one so I can properly assign to you.

k1s:
@lresende created https://issues.apache.org/jira/secure/ViewProfile.jspa?name=k1s

PR #151, opened 2018-01-24
[TOREE-463] break out scala interpreter initialization to speed up startup

brad-kaiser:
The goal of this PR is to decrease Toree's startup time by breaking out the ScalaInterpreter initialization and running it in its own thread. If you count the startup time as the time it takes KernelBootstrap.initizalize to complete, I saw the average startup time go from 8.16s to 5.69s on my laptop. I started the kernel ten times with the original code and my changes.

### DIFF
--- a/kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
+++ b/kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
@@ -28,6 +28,7 @@ import org.apache.toree.comm.{CommManager, CommRegistrar, CommStorage, KernelCom
 import org.apache.toree.dependencies.{CoursierDependencyDownloader, Credentials, DependencyDownloader}
 import org.apache.toree.interpreter._
 import org.apache.toree.kernel.api.Kernel
+import org.apache.toree.kernel.interpreter.scala.ScalaInterpreter
 import org.apache.toree.kernel.protocol.v5.KMBuilder
 import org.apache.toree.kernel.protocol.v5.kernel.ActorLoader
 import org.apache.toree.magic.MagicManager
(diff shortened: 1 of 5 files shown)


kernel/src/main/scala/org/apache/toree/boot/layer/ComponentInitialization.scala
```diff
@@ -28,6 +28,7 @@ import org.apache.toree.comm.{CommManager, CommRegistrar, CommStorage, KernelCom
 import org.apache.toree.dependencies.{CoursierDependencyDownloader, Credentials, DependencyDownloader}
 import org.apache.toree.interpreter._
 import org.apache.toree.kernel.api.Kernel
+import org.apache.toree.kernel.interpreter.scala.ScalaInterpreter
```
sanjay-saxena: Based on my limited understanding of the code, picking up a dependency on an implementation such as `ScalaInterpreter` from `scala-interpreter` project in the generic `kernel` project maybe an abstraction violation. It may also lead to circular dependencies. Maybe, you should look at either reusing the existing APIs on the generic trait `Interpreter`. And, if the existing APIs are not sufficient, then maybe add new APIs to the generic trait `Interpreter`.

kernel/src/main/scala/org/apache/toree/boot/layer/InterpreterManager.scala
```diff
@@ -20,15 +20,20 @@ package org.apache.toree.boot.layer
 import org.apache.toree.kernel.api.KernelLike
 import com.typesafe.config.Config
 import org.apache.toree.interpreter._
-import scala.collection.JavaConverters._
+import org.apache.toree.kernel.interpreter.scala.ScalaInterpreter
```
sanjay-saxena: Same comment about abstraction violation applies here as well.

PR #162, opened 2018-09-20
[TOREE-474] support for custom resolvers (for %AddDeps)

rmcentyre:
The code path where `downloadTotal` is set is inconsistently reached for certain repositories, like Artifactory. This should fix the masked issue (`java.util.NoSuchElementException: key not found` in `downloadTotal`) that was derailing the directory traversal process specified in TOREE-474. The visible error in the logs was the generic `An error occurred while downloading` specified in the Logger overrides.


lresende:
@rmcentyre do you have a JIRA account so I can assign the JIRA to your username? Otherwise, do you want to create one?

rmcentyre:
@lresende Just created. Same as the username above.

PR #167, opened 2018-11-23
[TOREE-492] Add DataFrame displayer

manuzhang:


manuzhang:
@lresende could you please take a look ?

lresende:
- What's the main difference between this and %dataframe
- Do we have to do any special handle for dataset ?
- Could you please add a little user documentation on how to use it (other displayers might be helpful as well).
- Could you please add a simple test case

manuzhang:
@lresende 
1. I was not aware of the `%dataframe` magic. This PR adds a syntactic sugar over `df.head(20)` where both `df` and `df.head` need `%truncation off` to be displayed in `html` format. In practice, I do a lot of `df.show()`s to get sample results which are not pretty printed. Now I know it can be achieved through `%dataframe` but I wonder it would better to display sampled `df` in `html` format by default (without any magic).

2. Do you know where to add the corresponding test case and doc for `ScalaDisplayer` ? All I've found are some notebook tutorials.

mohammedi-haroune:
@manuzhang It would be great if the displayer works by default (the same way as a pandas data frame). Is that working for you ?

manuzhang:
@mohammedi-haroune yes, that works for me

mohammedi-haroune:
@manuzhang Thanks for you reactivity. I will check it out.

lresende:
For now, I am inclined to continue the support for %%dataframe  as it provides more flexibility, additional controls and other format supported. Maybe we could build the additional flexibility and truncation indicators to make this more flexible then we can evaluate again

```
┌────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                │                                       PR #167 displayer                                        │                                          %%dataframe magic                                           │
├────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Trigger        │ Automatic — any DataFrame as last cell result                                                  │ Explicit %%dataframe [args] cell magic, code on following lines                                      │
├────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Output formats │ HTML only (fixed)                                                                              │ html, csv, json (--output flag)                                                                      │
├────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Row limit      │ Hardcoded 20 (df.head(20))                                                                     │ Configurable via --limit (default 10)                                                                │
├────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Rendering      │ Reuses displayRows (same table helper used for Array[Row]/Spark shell-style output)            │ Own HTML/CSV/JSON builders in DataFrameConverter                                                     │
├────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Error handling │ None — relies on default exception propagation                                                 │ Structured: distinguishes Aborted/Error/Incomplete interpreter results, returns usage/help on bad    │
│                │                                                                                                │ args                                                                                                 │
├────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Scope          │ Scala interpreter only (ScalaDisplayers)                                                       │ Any interpreter registered with the kernel (goes through kernelInterpreter.interpret)                │
├────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Value          │ Truncates silently at 20 rows, no truncation indicator (unlike Array[Row]'s isTruncated        │ Limit is explicit and user-controlled                                                                │
│ semantics      │ support)                                                                                       │                                                                                                      │
└────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

PR #170, opened 2019-06-13
Location for user dependencies +

Freshwood:
We had some problems with the nexus repository in conjunction with the Apache Toree incubator.

The coursier dependency manager is incompatible with Nexus repository manager. We could not use the magics %AddJar + %AddDeps in a normal way.

With the Ivy dependency manager we had no problems.

Changes:

We introduced a new setting 'ivy-settings' where an user can link an ivy-configuration file
Maybe it is useful for other people, too.


lresende:
@Freshwood Could you please describe a little bit more on the changes related to the `location for user dependencies`.

PR #179, opened 2019-10-28
Adding support for Scala 2.12 and Spark 2.4.4, plus test refactoring

Adi255:
To support using the Toree kernel with Spark 2.4.4+ I had to patch in support for Scala 2.12. This works for me now, so thought I would try merging it back.


Makefile
```diff
@@ -27,13 +27,13 @@ IS_SNAPSHOT?=true
...
-APACHE_SPARK_VERSION?=2.2.2
-SCALA_VERSION?=2.11
+APACHE_SPARK_VERSION?=2.4.4
+SCALA_VERSION?=2.12
 IMAGE?=jupyter/all-spark-notebook:228ae7a44e0c
 EXAMPLE_IMAGE?=apache/toree-examples
 SYSTEM_TEST_IMAGE?=apache/toree-systemtest
 GPG?=gpg
-GPG_PASSWORD?=
+GPG_PASSWORD?=sl
```
lresende: I believe this should be left empty

client/src/test/scala/examples/ScalaSparkClientUsage.scala
```diff
@@ -142,7 +142,7 @@ object ScalaSparkClientUsage extends App {
     y <- yPromise.future
   } yield (complexMath(x, y))
 
-  resultPromise.onSuccess {case x => println(s"Added result is ${x}") }
+  resultPromise.onComplete { x => println(s"Added result is ${x}") }
```
lresende: why this change? is this related to 2.12?

plugins/src/test/scala-2.12/org/apache/toree/plugins/dependencies/ClassLoaderHelper.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.apache.toree.plugins.dependencies
+
+import java.net.URL
+import java.lang.{ ClassLoader => JClassLoader }
+
+
+/**
+  * Created by mariu on 2016-07-16.
+  */
```
lresende: Please remove these comments...

lresende:
What is the issue you get today if you get branch and just build with `scala 2.12`? 
Also, we can't move and make toree always produce only `2.12` artifacts, I would recommend publishing both `2.12` and `2.11`.

Makefile
```diff
@@ -27,13 +27,13 @@ IS_SNAPSHOT?=true
 SNAPSHOT:=-SNAPSHOT
 endif
 
-APACHE_SPARK_VERSION?=2.2.2
-SCALA_VERSION?=2.11
+APACHE_SPARK_VERSION?=2.4.4
+SCALA_VERSION?=2.12
```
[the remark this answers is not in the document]
Adi255: That's reasonable; will change

Makefile
Adi255: Must be a typo on my part, good spot

client/src/test/scala/examples/ScalaSparkClientUsage.scala
Adi255: Indeed, onSuccess is deprecated in Scala 2.12 whereas onComplete is fine for both 2.11/12

plugins/src/test/scala-2.12/org/apache/toree/plugins/dependencies/ClassLoaderHelper.scala
Adi255: Will do

f1yegor:
@Adi255 do you have 2.12 version prebuild somewhere that I could reuse it either as pip package, or docker image?

Adi255:
Hi,
I do have a pip-installable tgz file, but it's 26 MB so I'm not sure how
best to get it to you? Please suggest something if you have any ideas. I
suppose you ought to be able to generate it yourself from the Makefile as I
did?

Thanks,
Adam

firestarman:
Any plan to support spark3.0 ? As you may know spark3.0-preivew2 has already been released.

Bidek56:
When can I expect a release that includes this PR? Thx

PR #180, opened 2019-12-13
[TOREE-508] Reply errors to iopub.error instead of iopub.exec_result

gmcoringa:
Fix string values within Metadata, the values needs to be json string.
Reply errors to iopub.error instead of iopub.exec_result

### DIFF
--- a/kernel/src/main/scala/org/apache/toree/kernel/protocol/v5/handler/ExecuteRequestHandler.scala
+++ b/kernel/src/main/scala/org/apache/toree/kernel/protocol/v5/handler/ExecuteRequestHandler.scala
@@ -89,19 +89,29 @@ class ExecuteRequestHandler(
         case Success(tuple) =>
           val (executeReply, executeResult) = updateCount(tuple, executionCount)
 
-          //  Send an ExecuteReply to the client
-          val executeReplyMsg = skeletonBuilder
-            .withHeader(MessageType.Outgoing.ExecuteReply)
-            .withContentString(executeReply).build
-          relayMsg(executeReplyMsg, relayActor)
+          if (executeReply.status.equals("error")) {
+            // Send an ExecuteReplyError with the result of the code execution to ioPub.error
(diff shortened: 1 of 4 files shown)


lresende:
I believe we are going in the right direction here, but with the change on behavior the errors are not completely showing up on the notebook interface.

Original behavior:
toree-508-original

And with the changes from this pr:
toree-508-updated

As shown in the picture above, the contents are not all there, and the cell is not scrollable anymore.

gmcoringa:
> I believe we are going in the right direction here, but with the change on behavior the errors are not completely showing up on the notebook interface.
> 
> Original behavior:
> toree-508-original
> 
> And with the changes from this pr:
> toree-508-updated
> 
> As shown in the picture above, the contents are not all there, and the cell is not scrollable anymore.

The error is not completely showing because of the code below found in [ScalaInterpreterSpecific.scala#L429](https://github.com/apache/incubator-toree/blob/master/scala-interpreter/src/main/scala-2.11/org/apache/toree/kernel/interpreter/scala/ScalaInterpreterSpecific.scala#L429):
```java
        ExecuteError(
          ex.getClass.getName,
          ex.getLocalizedMessage,
          formattedException.slice(1, formattedException.size - 1).toList
        )
```
Which removes the first line from the stack, I don't know why, but I believe that this should not be done.

About the scrollable, testing with python kernels seems to have the same behaviour (errors are not scrollable).

So @lresende if you think its better I can fix the removal of messages from the stack trace.

lresende:
@gmcoringa Yes, please update with the fix for the removal of the message.

gmcoringa:
@lresende fix done, also I did another commit to fix some unit tests, but there are other testing failing, that I couldn't understand why.

gmcoringa:
> @lresende fix done, also I did another commit to fix some unit tests, but there are other tests failing, that I couldn't understand why.

Another point is that I was not able to reproduce the error.

PR #191, opened 2020-09-24
Update README.md

sunbuhui:
I think this should be consistent with the installation command on the official website, because not everyone knows where spark is installed. Modified this block, we can remind users to specify the spark path
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,20 +96,22 @@ Refer to your `Docker` setup for the ip address. The notebook will be at `http:/
 
 Install
 =======
-
-Dev snapshots of Toree are located at https://dist.apache.org/repos/dist/dev/incubator/toree. To install using one
-of those packages, you can use the following:
-
+This requires you to have a distribution of Apache Spark downloaded to the system where Apache Toree will run. The following commands will install Apache Toree.
 ```
 pip install --upgrade toree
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,8 +101,8 @@ Dev snapshots of Toree are located at https://dist.apache.org/repos/dist/dev/inc
 of those packages, you can use the following:
 
 ```
-pip install <PIP_RELEASE_URL>
-jupyter toree install
+pip install --upgrade toree
+jupyter toree install --spark_home=/usr/local/bin/apache-spark/
 ```
 
 where `PIP_RELEASE_URL` is one of the `pip` packages. For example:
```
lresende, line 108: this paragraph will now become obsolete.

Key2Love: ok,but when i enter the command "pip install <PIP_RELEASE_URL> 
jupyter toree install", jupyter console reports "spark not found" cause jupyter sets spark_home as default,actually i didn't install spark at that path. I think it is a little bit ambiguous

lresende: @Key2Love I am not disagreeing with your pr, per contrary, it's good, but with the change, we should remove the lines below as they were applying to old contents, no?

lresende (approved):

PR #199, opened 2023-02-13
Scala2.13 + Spark v3.3

requaos:


lresende:
Thanks for this @requaos , let me start some discussion on how to manage this with regard supporting for old Spark releases.

requaos:
I would appreciate some guidance there.

lresende:
So it looks like the tests didn't get updated. I spent a little time on it and got the scala tests updated, but still seeing some issues related to Mockito, let me try to fix those and then I will push a commit to your pr.

lresende:
I finally found some time to spend on this PR, it's not completely passing all tests, but we are down to a hand full ones to fix.

lresende:
For testing it locally, instead of just `make dist` try `make clean dist test`

ywskycn:
@lresende any update on this change?

pan3793:
@requaos @lresende would you mind me taking over this Spark & Scala upgrading work? I would like to split this into some small tasks to make it easy to review if you don't mind.

pan3793:
@requaos thanks for sharing the background, I skimmed through the change, the most of the efforts are spent on dependencies upgrading, and unfortunately, some of them introduce breaking changes so a lot of code need to be adjusted. e.g. Spark, scalatest, mockito, coursier. It is worth upgrading them in independent PRs.

And I believe some of the dependencies upgradings are not required, e.g. sbt, akka. Especially, akka changed the license from Apache V2 to BSL v1.1 in 2.7, the latter is not allowed by Apache projects.

https://www.thestack.technology/akka-license-change-lightbend-opensource/

pan3793:
@requaos May I ask a little more about the requirements of Scala 2.13?

This is a Spark ecosystem project, most dependencies are intended to align with Spark, as I mentioned above

> Spark 3.0 ~ 3.5 uses Scala 2.12 by default, it rarely uses a non-default Scala version in production

So I suppose changing it to Scala 2.13 and dropping 2.12 looks aggressive. It makes more sense if we can support both Scala 2.12 and 2.13 and use 2.12 in default, but this means introducing some refactoring of the building system.

pan3793:
@requaos Thanks, as part of this PR, I think you can upgrade coursier to the latest 2.x version in a separate PR first, to gain support for both Scala 2.12 and 2.13.

I would like to take a look at scalatest/mockito stuff after #201 gets merged.

pan3793:
Since https://github.com/apache/incubator-toree/pull/218 got merged, I will close this one.

PR #200, opened 2023-06-23
[TOREE-536] AddJar command to support Google cloud storage

ywskycn:


kristopherkane:
I have done this previously and tested it.  Although it does work in Dataproc to make Toree's internal download function to 'cache JARs', The jar classes are not useable in the notebook itself.  I assume not all permutations of Spark paths (runtime, repl, etc) are updated. 

You can test this in a the kernel by simply importing a class from a JAR that was called to download with this magic.

ywskycn:
> I have done this previously and tested it. Although it does work in Dataproc to make Toree's internal download function to 'cache JARs', The jar classes are not useable in the notebook itself. I assume not all permutations of Spark paths (runtime, repl, etc) are updated.
> 
> You can test this in a the kernel by simply importing a class from a JAR that was called to download with this magic.

@kristopherkane could you tell more about "not usable" here? I tested by `AddJar` and then `import` classes from that jar in the notebook, and it works there. Also I verified in the Spark application UI, where the jar has been added to the classpath.

kristopherkane:
@ywskycn Interesting. I compiled this same change last week and did that to no avail on Dataproc 2.1. Is Dataproc also your runtime environment?

ywskycn:
@kristopherkane yes. What I have done:

- Apply this change to tags/v0.5.0-incubating-rc5 and `make dist`
- Use the newly built `dist/toree/lib/toree-assembly-0.5.0-incubating.jar` to replace `/opt/conda/miniconda3/lib/python3.10/site-packages/toree/lib/toree-assembly-0.5.0-incubating.jar` and `/opt/conda/miniconda3/share/jupyter/kernels/apache_toree_scala/lib/toree-assembly-0.5.0-incubating.jar` in a Dataproc cluster (image 2.1-debian11 with Jupyter component enabled).
- Test %AddJar and import command in a new notebook file, and it works well there.

kristopherkane:
Those are my exact steps and toree jar locations.  

Did you by chance change any spark:spark* configuration at cluster creation?

kristopherkane:
I also should say, my comment doesn't hold this PR up.  That schema will be required regardless.

ywskycn:
> Those are my exact steps and toree jar locations.
> 
> Did you by chance change any spark:spark* configuration at cluster creation?

no, I just follow all default configs, only specifying image version and Jupyter component.

kristopherkane:
OK, thanks!

lresende (approved):

almaaouiwassim:
Hello, it seems I have the same issue as you @kristopherkane even with the local `file` prefix. 

Screenshot 2023-08-11 at 10 38 15


I am using Dataproc 2.1, by defaut I have one driver and two executors and the yarn deploy mode is `client`. Seing the jar load messages duplicated twice, I am wondering if the classes are being loaded on the executors but not on the driver?

kristopherkane:
@almaaouiwassim sorry, I did not.  I've kind-of given up on Toree since we will also need kernel time Spark configuration changes and the level of effort to add those is too high, in addition to the driver not getting the AddJar classpath.

lresende:
Could you clarify your additional requirements?

kristopherkane:
@lresende The ability to set Spark configuration from the kernel itself and before the SparkSession is created.

almaaouiwassim:
Hey guys, I am back with few steps to reproduce the issue. The `AddJar` seems to work with some jars but no all of them, I couldn't get what makes my jar not well loaded by the kernel. 

Here are the steps to reproduce the issue

- create a 2.1.21-debian11 dataproc cluster (probably the jar won't work with other env, but this is the one I am testing with)

- download the jar to the local filesystem using terminal on the notebook ```gsutil cp gs://untienots_public/myproject-assembly.jar .```

- load the jar in the notebook then do an import 

You can check the screenshots to see the results and to see that my jar actually contains the class that couldn't be imported

Screenshot 2023-08-16 at 14 10 45
Screenshot 2023-08-16 at 14 11 04

ywskycn:
@almaaouiwassim have you checked the Spark UI whether the jar was added?

almaaouiwassim:
@ywskycn yes it has been added to the spark configuration
Screenshot 2023-08-17 at 10 35 38

almaaouiwassim:
@ywskycn did you succeed to reproduce the issue please?

ranadip:
I have been able to reproduce the same issue. It seems the jar in the example @ywskycn suggested works fine with AddJars but one created locally using sbt does not load even though it gets downloaded into the local Toree cache from GCS. No errors/warnings show up either on screen or in the logs.

pan3793:
image

Can not reproduce the issue with the latest master [commit](https://github.com/apache/incubator-toree/commit/937c1260bf7973ef20b0052989398ac9e789d160)

pan3793:
@almaaouiwassim I'm not familiar with the GCP platform, I suppose the jar size is small, would you mind sending it to me via email?

> I suspect it's linked to how the jar is actually built. If so I can provide my `build.sbt` file

I believe it would be helpful if you could share how to produce your problematic(maybe) jar from scratch.

almaaouiwassim:
Hello guys, sorry for the late response, I had time recently to try to understand what exactly doesn't work with Apache Toree. It seems that any jar with a class on the root classpath of the jar (not under a package) is not well added to the kernel using AddJar. 

I put a sample project to reproduce it in https://github.com/almaaouiwassim/my_project. just run `sbt assembly` to generate the jar. 

I have also put the jar on GCS https://storage.cloud.google.com/untienots_public/my_project.jar?authuser=0


Please tell me if you need further information. 

FI, even if you don't put classes at the root classpath, some dependencies can still do it and break your jar, in our case it was `spark-xml` who has a class `module-info.class` at the root. I guess many dependencies would do the same.

PR #202, opened 2023-08-05
[TOREE-526] Enable system test

pan3793:
Enables `make system-test` in CI so `jupyter_kernel_test` runs on Java 8/11 and Scala 2.12/2.13.

The `build` matrix is green on all four jobs and the step reports `Ran 14 tests ... OK (skipped=9)`. The skipped cases are the optional test groups with no samples defined in `test_toree.py`.

**Known flake**
`test_scala_execute_result` asserts on the last iopub message. Cells that run a magic produce both a `stream` message (the magic's output) and an `execute_result`, so that assertion can pick the stream message and fail with `KeyError: 'data'`. Observed once in CI, not reproducible in 30 local 2.13 runs; asserting on the last `execute_result` would remove the sensitivity.


pan3793:
https://github.com/apache/incubator-toree/pull/207 is opened to address "status" issue.

scala-interpreter/src/main/scala-2.13/org/apache/toree/kernel/interpreter/scala/ScalaInterpreterSpecific.scala
```diff
@@ -309,7 +309,7 @@ trait ScalaInterpreterSpecific extends SettingsProducerLike { this: ScalaInterpr
     logger.debug(s"Attempting code completion for ${code}")
     val result = completer.complete(code, pos)
 
-    (result.cursor, result.candidates.map(_.toString))
+    (result.cursor, result.candidates.map(_.name))
```
pan3793, line 312: Only the 2.13 line needs this: the 2.12 REPL (`PresentationCompilerCompleter`) returns `Completion.Candidates(cursor, List[String])`, so `candidates` is already a list of names and `CompletionCandidate` does not exist in the 2.12 compiler jar. The 2.13 REPL returns `List[CompletionCandidate]`, hence the mapping to `name` here.

pan3793:
@lresende this one is ready and fixed a real bug for scala 2.13, let me know whether you want to include it in 0.6.0

lresende (approved):

PR #203, opened 2023-08-05
[TOREE-538] apache/toree-dev image supports Applie Silicon

pan3793:
I got failures when running `make dev` on M1 chips MacBook
```
29.95 update-alternatives: error: alternative /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java for java not registered; not setting
```

because for ARM platform, the installed OpenJDK 8 path is `/usr/lib/jvm/java-8-openjdk-arm64` instead of `/usr/lib/jvm/java-8-openjdk-amd64`.

I verified this change locally by run `make dev` on my M1 MacBook, and the CI should verify the change works on x86 platform.

### DIFF
--- a/Dockerfile.toree-dev
+++ b/Dockerfile.toree-dev
@@ -30,7 +30,8 @@ RUN apt-get -y update && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/* && \
     update-ca-certificates -f && \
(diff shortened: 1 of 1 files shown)


pan3793:
The UT stuck is irrelevant, fixed in TOREE-540 (https://github.com/apache/incubator-toree/pull/205)

lresende (approved):

PR #207, opened 2023-08-06
[TOREE-541] Reply message should implement status field

pan3793:
This PR refers to https://jupyter-client.readthedocs.io/en/latest/messaging.html to revise the Jupyter message protocol implementation

> All reply messages have a 'status' field ...

A trait `ReplyContent` is introduced to make sure all reply messages meet this contract.

One exception is "connect_reply", because

> Deprecated since version 5.1: connect_request/reply have not proved useful, and are considered deprecated. Kernels are not expected to implement handlers for this message.

https://jupyter-client.readthedocs.io/en/latest/messaging.html#connect

### DIFF
--- a/client/src/main/scala/org/apache/toree/kernel/protocol/v5/client/socket/StdinClient.scala
+++ b/client/src/main/scala/org/apache/toree/kernel/protocol/v5/client/socket/StdinClient.scala
@@ -22,7 +22,7 @@ import org.apache.toree.communication.ZMQMessage
 import org.apache.toree.communication.security.SecurityActorType
 import org.apache.toree.kernel.protocol.v5.client.ActorLoader
 import org.apache.toree.kernel.protocol.v5.{HeaderBuilder, KMBuilder, KernelMessage}
-import org.apache.toree.kernel.protocol.v5.content.{InputReply, InputRequest}
+import org.apache.toree.kernel.protocol.v5.content.{InputReply, InputReplyOk, InputRequest}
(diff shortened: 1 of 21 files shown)


pan3793:
kindly ping @lresende and @kevin-bates, would you mind taking a look when you have time?

lresende:
I see a lot of updates adding status "ok"... how about error conditions when something fail?

pan3793:
@lresende I think the code won't reach there if something fails before it. Note that, only a few kind simple reply messages did not implement the status, for message like `ExecuteReply` which already has status field with properly handled.

pan3793:
@lresende FYI, I changed the code to use `type XyzReplyOk = XyzReply` to keep consistent with the current code, no functionality changed.

pan3793:
@lresende @kevin-bates would you mind taking another look? IMO this does not change the state of error handling, should be safe.

lresende:
Have you tested this with real job notebooks on latest jupyterlab 3.x and no side effects?

pan3793:
@lresende I don't have JuypterLab environment, instead, I tested it with built-in `examples/magic-tutorial.ipynb`, everything goes well

image

pan3793:
Kindly ping @lresende @kevin-bates

pan3793:
@lresende I rebased it on the latest master and fixed the conflict, please take a look when you have time.

lresende (approved):

PR #211, opened 2023-08-10
[TOREE-545] Upgrade Scalatest 3.2 and Mockito 4

pan3793:
This is part of https://github.com/apache/incubator-toree/pull/199

Co-authored-by: Luciano Resende <lresende@apple.com>
Co-authored-by: Cheng Pan <chengpan@apache.org>

This PR aims to upgrade scalatest to 3.2 and the corresponding mockito to 4.11 to prepare support for Scala 2.13

The change is verified by `examples/magic-tutorial.ipynb`

image

### DIFF
--- a/client/src/test/scala/org/apache/toree/kernel/protocol/v5/client/socket/IOPubClientSpec.scala
+++ b/client/src/test/scala/org/apache/toree/kernel/protocol/v5/client/socket/IOPubClientSpec.scala
@@ -32,12 +32,14 @@ import org.apache.toree.kernel.protocol.v5.client.execution.{DeferredExecution, 
 import org.apache.toree.kernel.protocol.v5.client.{ActorLoader, Utilities}
 import org.apache.toree.kernel.protocol.v5.content.{CommClose, CommMsg, CommOpen, StreamContent}
 import com.typesafe.config.ConfigFactory
-import org.mockito.Matchers.{eq => mockEq, _}
+import org.mockito.ArgumentMatchers.{eq => mockEq, _}
 import org.mockito.Mockito._
 import org.scalatest.concurrent.{Eventually, ScalaFutures}
 import org.scalatestplus.mockito.MockitoSugar
 import org.scalatest.time.{Milliseconds, Span}
-import org.scalatest.{BeforeAndAfter, FunSpecLike, Matchers}
+import org.scalatest.funspec.AnyFunSpecLike
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.BeforeAndAfterEach
 import play.api.libs.json.Json
 
 import scala.concurrent.duration._
@@ -53,8 +55,8 @@ object IOPubClientSpec {
 
 class IOPubClientSpec extends TestKit(ActorSystem(
   "IOPubClientSpecSystem", ConfigFactory.parseString(IOPubClientSpec.config)
-)) with ImplicitSender with FunSpecLike with Matchers with MockitoSugar
-  with ScalaFutures with BeforeAndAfter with Eventually
+)) with ImplicitSender with AnyFunSpecLike with Matchers with MockitoSugar
+  with ScalaFutures with BeforeAndAfterEach with Eventually
 {
   private val TestTimeout = Timeout(10.seconds)
   implicit override val patienceConfig = PatienceConfig(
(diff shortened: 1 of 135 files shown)


client/src/test/scala/org/apache/toree/kernel/protocol/v5/client/socket/IOPubClientSpec.scala
```diff
@@ -108,7 +110,7 @@ class IOPubClientSpec extends TestKit(ActorSystem(
           .build
 
         // Mark as target being provided
-        doReturn(Some(mockCommCallbacks)).when(spyCommStorage)
+        doReturn(Some(mockCommCallbacks), Nil: _*).when(spyCommStorage)
```
pan3793, line 113: This is a Scala bug, fixed in 2.13.0. A workaround is required to make it happy with Scala 2.12.

https://github.com/scala/scala/pull/7680

pan3793:
@lresende would you mind taking a look at this PR? it is based on your previous in-progress work.

lresende:
@pan3793 I believe this is mostly done in #199, and I was wondering what would be best here. Contribute things in pieces because #199 updates scala do 2.13? Or should we continue adding to that PR to fix the remaining few test issues and merge #199 pr?

pan3793:
@lresende this PR is ready to go(review and merge) on my side.

> I was wondering what would be best here.

In short words, I intend to make each PR focus on one goal, especially, the large goal should be broken down into small/middle ones, which are easy to implement and review.

For #199, technically, I think it is still far from the goal, because

1. I and @requaos reached a consensus that we should not drop Scala 2.12 in Spark 3.x periods, as I explained before, Spark 3.x uses  Scala 2.12 in default, it rarely uses the non-default Scala version in production.
   >  (from @requaos) Planning forward to support both Scala 2.12 and 2.13, with a default preference for 2.12, will be most optimal.

   #199's approach totally breaks the Scala 2.12 support.

2. coursier upgrading with some breaking changes, and the #199's changed the logging behavior. It is also worth accomplishing in an independent PR.

3. #199 contains many irrelevant changes, e.g. it upgraded Java 8 to 11, and it's hard to thoroughly review with so many lines of change.

As @requaos said, I think he also agrees to break it down.

> ... this is a larger effort than initially planned ...

> Please let me know what I can do to prepare this PR into the singular context of Scala 2.13 support.

> I can accommodate upstream changes in my fork to assist in this effort. As you see fit to divide this labor, I will gladly reduce the scope of my PR to reflect an incremental task to work more gradually towards our shared goal.

@lresende That's my opinion, my goal is to accomplish the upgrading quickly, keep code quality, and make the code change history clear. Please let me know if you have any other thoughts :)

project/Dependencies.scala
```diff
@@ -45,15 +45,15 @@ object Dependencies {
...
   val playJson = "com.typesafe.play" %% "play-json" % "2.7.4" // Apache v2
 
   val scalaCompiler = Def.setting{ "org.scala-lang" % "scala-compiler" % scalaVersion.value } // BSD 3-clause
   val scalaLibrary = Def.setting{ "org.scala-lang" % "scala-library" % scalaVersion.value } // BSD 3-clause
   val scalaReflect = Def.setting{ "org.scala-lang" % "scala-reflect" % scalaVersion.value } // BSD 3-clause
 
-  val scalaTest = "org.scalatest" %% "scalatest" % "3.0.8" // Apache v2
+  val scalaTest = "org.scalatest" %% "scalatest" % "3.2.16" // Apache v2
+  val scalaTestMockito = "org.scalatestplus" %% "mockito-4-11" % "3.2.16.0" // Apache v2
+  val mockitoInline = "org.mockito" % "mockito-inline" % "4.11.0" // MIT
```
pan3793, line 56: all are test only, so no need to update LICENSE/NOTICE

pan3793: BTW, `mockito-inline` is introduced to support mock final class / enum

pan3793:
The test failure is irrelevant, it's another known flaky test, will investigate independently.

```
[info]   - should send a status idle message after the reply and result *** FAILED *** (15 seconds, 40 milliseconds)
[info]     0 did not equal 2 (ExecuteRequestHandlerSpec.scala:193)
[info]     org.scalatest.exceptions.TestFailedException:
[info]     at org.scalatest.matchers.MatchersHelper$.indicateFailure(MatchersHelper.scala:397)
[info]     at org.scalatest.matchers.should.Matchers$ShouldMethodHelperClass.shouldMatcher(Matchers.scala:7299)
[info]     at org.scalatest.matchers.should.Matchers$AnyShouldWrapper.should(Matchers.scala:7362)
[info]     at org.apache.toree.kernel.protocol.v5.handler.ExecuteRequestHandlerSpec.$anonfun$new$6(ExecuteRequestHandlerSpec.scala:193)
```

pan3793:
@lresende opened #218(I have to open a new PR because I don't have permission to push commit the original PR branch), which is a rough and dirty fix of #199, it passed building and testing on both Scala 2.12 and 2.13.

pan3793:
@lresende do you have further concerns about shipping this patch?

pan3793:
This PR only touches the test code, since all tests passed and no further objections, I'm going to merge it to unblock Scala 2.13 upgrading tasks.

PR #216, opened 2023-08-11
Revert "[TOREE-543] Fix and enable JVMReprSpec"

pan3793:
Reverts apache/incubator-toree#209

It fixes the UT, but unfortunately, breaks the real workload ...

image

We need to explore another way to adjust the test case, or change the implementation.


pan3793:
@lresende I'm really sorry about that, will do such testing for other PRs in the future.

pan3793:
@lresende thanks!

PR #219, opened 2023-08-23
[Toree-551] Always populate traceback in reply error.

functicons:


pan3793:
Seems the master branch works well without this change.

image

image

pan3793:
Same result in Jupyter NbClassic

image

image

pan3793:
@functicons maybe this one https://github.com/jupyter/notebook/pull/6466

PR #194, opened 2021-11-04
adjustments for Java 11 and Spark 3.2.0

pikulmar:

PR #220, opened 2023-08-29
[TOREE-552] Downgrade Guava 14.0.1 to align with Spark

pan3793:
### What changes were proposed in this pull request?

This PR proposes to downgrade Guava 14.0.1 to align with Apache Spark.

The previous Guava upgrade is to make the `ClassPath` compatible with Java 11, https://github.com/google/guava/issues/3249. As a workaround, this PR fork a `ClassPath` from Guava v32.1.2 to make it happy with Java 11.

### Why are the changes needed?

Upgrading Guava in Spark is a "Mission Impossible" at the current stage. See details at SPARK-44811 SPARK-36676

The current tests do not cover the use cases of Spark with Hive, and based on the above Spark tickets, the Guava upgrading would break Hive usage

### DIFF
--- a/kernel/src/main/scala/org/apache/toree/utils/ClassPath.java
+++ b/kernel/src/main/scala/org/apache/toree/utils/ClassPath.java
@@ -14,6 +14,9 @@
  *  See the License for the specific language governing permissions and
  *  limitations under the License
  */
+
+// Below are original header from Guava project
+
 /*
  * Copyright (C) 2012 The Guava Authors
  *
--- a/kernel/src/main/scala/org/apache/toree/utils/ClassPath.java
+++ b/kernel/src/main/scala/org/apache/toree/utils/ClassPath.java
@@ -53,8 +53,12 @@ import java.util.jar.Manifest;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 
-// Copied from Guava v32.1.2
+// This class is copied from Guava v32.1.2
 // https://github.com/google/guava/blob/v32.1.2/guava/src/com/google/common/reflect/ClassPath.java
+//
+// TOREE-552: Apache Toree should use Guava 14 to align with Apache Spark, unfortunately, ClassPath
+// from Guava 14 does not work well with Java 11+.
+// See detail at: https://github.com/google/guava/issues/3249
 
 /**
  * Scans the source of a {@link ClassLoader} and finds all loadable classes and resources.
--- a/kernel/src/main/scala/org/apache/toree/utils/ClassPath.java
+++ b/kernel/src/main/scala/org/apache/toree/utils/ClassPath.java
@@ -1,0 +1,687 @@
+/*
+ * Copyright (C) 2012 The Guava Authors
+ *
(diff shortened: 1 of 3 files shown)


project/Dependencies.scala
```diff
@@ -67,24 +67,13 @@ object Dependencies {
...
-  val guava = "com.google.guava" % "guava" % "31.0.1-jre" // Apache v2
+  // TOREE-552 Align Guava with Spark
+  val guava = "com.google.guava" % "guava" % "14.0.1" // Apache v2
 
   // Projects
-
   val sparkAll = Def.setting{
     Seq(
-      sparkCore.value % "provided" excludeAll(
-        // Exclude netty (org.jboss.netty is for 3.2.2.Final only)
```
pan3793, line 82: The recent version of Spark does not depend on Netty 3.x, so this exclusion is redundant.

https://mvnrepository.com/artifact/org.apache.spark/spark-core_2.12/3.3.2

pan3793:
CI reports failures, investing

kernel/src/main/scala/org/apache/toree/utils/ClassPath.java
```diff
@@ -0,0 +1,720 @@
...
+import java.util.jar.JarFile;
+import java.util.jar.Manifest;
+import javax.annotation.CheckForNull;
+
+// This class is derived from Guava v32.1.2
+// https://github.com/google/guava/blob/v32.1.2/guava/src/com/google/common/reflect/ClassPath.java
+//
+// TOREE-552: Apache Toree should use Guava 14 to align with Apache Spark, unfortunately, ClassPath
+// from Guava 14 does not work well with Java 11+.
+// See detail at: https://github.com/google/guava/issues/3249
```
pan3793, line 78: This is a simple copy with limited changes:

1. switch `Logger` from JDK logging to SFL4J
2. directly call `System.getProperty` to get `path.separator` and `java.class.path` instead of using `com.google.common.base.StandardSystemProperty`(since Guava 15)

pan3793: This looks like an exception because it is the first Java file in the project, though sbt works well with Java source files, if requested, I can translate it to Scala.

pan3793:
@lresende would you mind taking a look?

lresende:
So, #194 updated Guava, and why now, when we downgrade to the same previous version, do we need some new workarounds? If Spark uses this, can't we import the additional files from Spark?

pan3793:
> when we downgrade to the same previous version, do we need some new workarounds?

Yes, I explained in the PR description. "As a workaround, this PR fork a `ClassPath` from Guava v32.1.2 to make it happy with Java 11."

> If Spark uses this, can't we import the additional files from Spark?

In short, Spark 3.x **must** use Guava 14.0.1, whose `ClassPath` is not compatible with Java 11. The restriction comes from the transitive deps of Spark, including Hive and Hadoop, see details in discussions in SPARK-44811 SPARK-36676

lresende (approved):

PR #227, opened 2025-02-18
fix issue with missing sbt in build.yml

pjfanning:
sbt is no longer added to `ubuntu-latest` in GitHub CI. You have to install it explicitly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,14 +41,16 @@ jobs:
       TEST_DILATION: 3
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
       with:
         clean: true
-    - name: Checkout
-      uses: actions/setup-java@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -41,14 +41,16 @@ jobs:
...
-    - name: Checkout
-      uses: actions/setup-java@v2
+    - name: Setup Java
+      uses: actions/setup-java@v4
       with:
         distribution: 'temurin'
         java-version: ${{ matrix.java }}
+        cache: sbt
+    - uses: sbt/setup-sbt@v1        
     - name: Add SBT launcher
```
pan3793, line 54: this step is useless then?

pan3793: I mean line 54-57

pan3793 (approved):

pan3793:
Thanks, merged to master

PR #229, opened 2025-09-05
Jupyter Spark monitor

vincentye38:
Implement a spark-monitor plugin to sent spark progress update to [SparkMonitor](https://krishnan-r.github.io/sparkmonitor/) frontend extension.
This plugin can be built as individual jar and enable by specifying  --magic-url" args[7] = "file:///path/to/spark-monitor-plugin.jar"

PR #282, opened 2026-09-17
Attribute SparkMonitor-derived code in spark-monitor-plugin

pjfanning:
## What

`spark-monitor-plugin/.../JupyterSparkMonitorListener.scala` (added in #229) is derived from SparkMonitor's [`scalalistener_spark3/CustomListener.scala`](https://github.com/swan-cern/sparkmonitor/blob/v3.3.0/scalalistener_spark3/CustomListener.scala) (Apache License 2.0). Roughly 60% of the file is verbatim upstream code, including the class/method doc comments. Neither `LICENSE` nor `NOTICE` mentioned SparkMonitor.

This PR:
- keeps the standard ASF header on the file and adds a note of origin below it, linking the upstream file and crediting the SWAN team at CERN — the same pattern used for the Guava-derived `ClassPath.java`;
- adds a `LICENSE` entry for the derived code next to the existing Guava entry.

`LICENSE` feeds `dist/toree-legal/LICENSE`, so the bin tarball, both assembly jars and both pip packages pick this up without further changes. Upstream ships no `NOTICE`, so nothing needs to propagate to ours.

## Verification

- `etc/tools/check-licenses` (RAT) passes.
- `sbt sparkMonitorPlugin/compile` succeeds.

Found while verifying 0.6.0-incubating RC6 (third-party provenance sweep of the source tarball).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


spark-monitor-plugin/src/main/scala/org/apache/toree/plugins/sparkmonitor/JupyterSparkMonitorListener.scala
```diff
@@ -15,6 +15,10 @@
  *  limitations under the License
  */
 
+// This file is derived from SparkMonitor's CustomListener.scala, originally written by the
+// SWAN team at CERN and licensed under the Apache License, Version 2.0.
+// https://github.com/swan-cern/sparkmonitor/blob/c12c0bc259eff9ad936971a616a08ab584138d10/scalalistener_spark3/CustomListener.scala
```
pan3793: use a tag?

pjfanning: Done - now points at the v3.3.0 tag (the file there is identical to master).

lresende (approved):

PR #283, opened 2026-09-18
Fix flaky size test in TaskManagerSpec

pan3793:
The `#size should reflect queued tasks and executing tasks` test adds two tasks and asserts a size of 2, but the second `add` can land before the first task starts executing, in which case the size is 1 and the assertion fails.

Wait for `isExecutingTask` before queueing the second task, the same way the `#size should be one if there is only one executing task` test already does.


pan3793:
@lresende this is a test stability enhancement, worth including in the next rc.

lresende (approved):

PR #284, opened 2026-09-18
Fork tests so Mockito's inline mock maker attaches per JVM

pan3793:
Mockito's inline mock maker self-attaches ByteBuddy's agent on first use, which on Java 8 loads `libattach.so` once per JVM. Tests that do not fork share the sbt JVM, so the first module to mock keeps the native library and every other module's test classloader fails:

```
java.lang.UnsatisfiedLinkError: Native Library .../libattach.so already loaded in another classloader
```

and the suite aborts with `MockitoInitializationException: Could not initialize inline Byte Buddy mock maker`. Observed on Java 8 + Scala 2.13, where a `communication` suite mocks before `kernel-api`'s do.

`ThisBuild / Test / fork := true` gives every module its own test JVM, which makes the per-module settings in `kernel`, `plugins` and `spark-monitor-plugin` redundant.


pan3793:
@lresende this is another test stability enhancement, worth including in the next rc.

lresende (approved):

PR #288, opened 2026-09-20
refer to Apache Toree (Incubating)

pjfanning:
ASF Incubator requires this


DISCLAIMER
```diff
@@ -1,4 +1,4 @@
-Apache Toree is an effort undergoing incubation at the Apache Software
+Apache Toree (Incubating) is an effort undergoing incubation at the Apache Software
```
lresende: Do we really need this change? The disclaimer is a template provided by the incubating PMC and disclaims right there that this is a project undergoing incubation? The issue is that we might have a couple of copies of this and we might need to validate all are properly in sync. 

https://incubator.apache.org/policy/incubation.html#standard_disclaimer

pjfanning: * the examples in use that I found has the Incubating in the name
* If I must remove this, I will
* but the others are needed

README.md
```diff
@@ -26,7 +26,7 @@
 
 [Apache Toree][website]
 ============
-Apache Toree is a [Juypter Notebook](https://jupyter.org/) kernel. The main goal of Toree is to provide the foundation for
+Apache Toree (Incubating) is a [Juypter Notebook](https://jupyter.org/) kernel. The main goal of Toree is to provide the foundation for
```
lresende, line 29: The policy states - you MUST include these statements in the README file of each repository... You SHOULD add a link to the Disclaimers in the README file. 

Maybe we should add a copy of the disclaimer or a link to the disclaimer at the bottom of the readme?

etc/templates/toree-assembly-pom.xml
```diff
@@ -39,7 +39,7 @@
     from the copy the ASF requires the distribution to carry.
   -->
   <description>
-    Apache Toree is a Jupyter Notebook kernel that provides interactive
+    Apache Toree (Incubating) is a Jupyter Notebook kernel that provides interactive
```
lresende, line 42: Is this required by policy? i was searching other podlings and from a handful, only one had this in their description.

lresende: and the Disclaimer is embedded to the description on the generated pom with @disclaimer@

pjfanning: As far as I can see, in practice, the name of the project is effectively 'Apache Toree (Incubating)' as opposed to Apache Toree or Toree.

pjfanning: To phrase this another way - on release votes and other reviews, you won't get -1s for saying Incubating too often but you can get -1s for omitting it.

README.md
pjfanning: added the disclaimer to the readme

DISCLAIMER
pjfanning: undid this change

etc/templates/toree-assembly-pom.xml
lresende: Won't disagree, but then we need to add to the other assembly (spark-monitor) ? For consistency?

pjfanning: I may not have time over the next few days or even this week to look at spark-monitor. Can we merge this as is though and get back to spark-monitor as a separate task?

pjfanning: undid this change due to incubation text being there

I got claude to investigate the spark monitor job and got it to fix up the pom there

I would still recommend that these files are verified before a release vote is started.

PR #289, opened 2026-09-20
[TOREE-576] Fix stale toree-assembly jar name in legal README

yew1eb:
The kernel-launch failure behind TOREE-576 was already fixed on master by 7f1a8438 (PR #286), which updated the glob in `etc/bin/run.sh`. Three example commands in `etc/legal/README.md` still referenced the pre-rename `target/scala-2.12/toree-assembly-*.jar` and now match nothing; point them at `apache-toree-assembly-*.jar`.


pan3793:
actually, we are discussing restoring the name
https://lists.apache.org/thread/spbg1z5yl5k4vpqvfgc7vopgg3cg9jwx

lresende:
As @pan3793 mentioned above, we are reverting these changes in #290

